PR #1, opened 2018-10-24
Initial implementation of d2l-sort-by

AdmiralCoco:
Updated usage: 
`<d2l-sort-by>`
- Attributes:
    - `value` (of thecurrently selected option)
    - `label` (for dropdown menu)
- Events:
    - Fires 'd2l-sort-by-option-selected’ 
        - { value: ‘…’ }

`<d2l-sort-by-option>`
- Attributes:
    - `text`
    - `value`
- Uses [`d2l-menu-item-radio-behavior`](https://github.com/BrightspaceUI/menu/blob/master/d2l-menu-item-radio-behavior.html)

---

Example usage:
```html
<d2l-sort-by-dropdown label="Sort by options">
  <d2l-sort-by-dropdown-option value="date" text="Date"></d2l-sort-by-dropdown-option>
  <d2l-sort-by-dropdown-option value="time" text="Time"></d2l-sort-by-dropdown-option>
</d2l-sort-by>
```

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,8 @@
     "wct.conf.json"
   ],
   "dependencies": {
-    "polymer": "Polymer/polymer#^2.0.0"
+    "polymer": "Polymer/polymer#^2.0.0",
+    "d2l-dropdown": "^6.6.1"
   },
   "devDependencies": {
     "d2l-typography": "BrightspaceUI/typography#^6.0.0",
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,7 @@
   ],
   "dependencies": {
     "polymer": "Polymer/polymer#^2.0.0",
-    "d2l-dropdown": "^6.6.1",
+    "d2l-dropdown": "^6.6.2",
     "d2l-menu": "^1.2.0",
     "d2l-icons": "^5.1.2"
   },
--- a/bower.json
+++ b/bower.json
@@ -17,8 +17,8 @@
   "dependencies": {
     "polymer": "Polymer/polymer#^2.0.0",
     "d2l-dropdown": "^6.6.2",
-    "d2l-menu": "^1.2.0",
-    "d2l-icons": "^5.1.2"
+    "d2l-icons": "^5.1.2",
+    "d2l-menu": "^1.2.0"
   },
   "devDependencies": {
     "d2l-typography": "BrightspaceUI/typography#^6.0.0",
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,9 @@
   ],
   "dependencies": {
     "polymer": "Polymer/polymer#^2.0.0",
-    "d2l-dropdown": "^6.6.1"
+    "d2l-dropdown": "^6.6.1",
+    "d2l-menu": "^1.2.0",
+    "d2l-icons": "^5.1.2"
   },
   "devDependencies": {
     "d2l-typography": "BrightspaceUI/typography#^6.0.0",
--- a/d2l-sort-by-dropdown-option.html
+++ b/d2l-sort-by-dropdown-option.html
@@ -23,7 +23,7 @@
 		 */
 		class SortByOption extends Polymer.mixinBehaviors(
 			[
-				D2L.PolymerBehaviors.MenuItemRadioBehavior,
+				D2L.PolymerBehaviors.MenuItemRadioBehavior
 			],
 			Polymer.Element
 		) {
@@ -31,10 +31,10 @@
 			static get properties() {
 				return {
 					text: {
-						type: String,
+						type: String
 					},
 					value: {
-						type: String,
+						type: String
 					}
 				};
 			}
--- a/d2l-sort-by-dropdown-option.html
+++ b/d2l-sort-by-dropdown-option.html
@@ -1,25 +1,14 @@
 <link rel="import" href="../polymer/polymer-element.html">
-<link rel="import" href="../d2l-dropdown/d2l-dropdown-menu.html">
-<link rel="import" href="../d2l-dropdown/d2l-dropdown-button-subtle.html">
-<link rel="import" href="../d2l-menu/d2l-menu-item.html">
-<link rel="import" href="../d2l-menu/d2l-menu-item-behavior.html">
+<link rel="import" href="../d2l-menu/d2l-menu-item-radio-behavior.html">
+<link rel="import" href="../d2l-menu/d2l-menu-item-selectable-styles.html">
+<link rel="import" href="../d2l-icons/d2l-icon.html">
+<link rel="import" href="../d2l-icons/tier1-icons.html">
 
 <dom-module id="d2l-sort-by-dropdown-option">
 	<template strip-whitespace>
-		<style>
-			:host {
-				display: flex;
-				flex-direction: column;
-			}
-			d2l-menu-item {
-				/**
-				 * Needed to show the background colour of d2l-menu-item when hovering;
-				 * d2l-menu-item is overriding the :hover style of d2l-menu
-				 */
-				background-color: transparent;
-			}
-		</style>
-		<d2l-menu-item text="[[text]]"></d2l-menu-item>
+		<style include="d2l-menu-item-selectable-styles"></style>
+		<d2l-icon icon="d2l-tier1:check" aria-hidden="true"></d2l-icon>
+		<span>[[text]]</span>
 	</template>
 	<script>
 		/**
@@ -28,7 +17,7 @@
 		 */
 		class SortByOption extends Polymer.mixinBehaviors(
 			[
-				D2L.PolymerBehaviors.MenuItemBehavior,
+				D2L.PolymerBehaviors.MenuItemRadioBehavior,
 			],
 			Polymer.Element
 		) {
@@ -36,6 +25,9 @@
 			static get properties() {
 				return {
 					text: {
+						type: String,
+					},
+					value: {
 						type: String,
 					}
 				};
--- a/d2l-sort-by-dropdown-option.html
+++ b/d2l-sort-by-dropdown-option.html
@@ -6,7 +6,13 @@
 
 <dom-module id="d2l-sort-by-dropdown-option">
 	<template strip-whitespace>
-		<style include="d2l-menu-item-selectable-styles"></style>
+		<style include="d2l-menu-item-selectable-styles">
+			/*
+			* https://github.com/Polymer/tools/issues/408
+			* Empty style blocks break linter.
+			*/
+			:host {}
+		</style>
 		<d2l-icon icon="d2l-tier1:check" aria-hidden="true"></d2l-icon>
 		<span>[[text]]</span>
 	</template>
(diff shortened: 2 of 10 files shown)


d2l-sort-by.html
```diff
@@ -0,0 +1,97 @@
...
+				this.$['d2l-sort-by-menu'].removeEventListener('d2l-menu-item-select',
+					this._boundListeners._onItemSelect);
+			}
+
+			_onItemSelect(e) {
+				this.selected = e.target.text;
+				this.dispatchEvent(new CustomEvent(
+					'd2l-sort-by-change',
+					{ bubbles: true, composed: true, detail: { value: this.selected } }
+				));
```
dlockhart: By using `<d2l-menu-item-radio>`, it fires its own events (see its documentation) that you could catch, stop propagation and then fire your own event, including the handy value.

d2l-sort-by.html
```diff
@@ -0,0 +1,97 @@
...
+				this.$['d2l-sort-by-menu'].close();
```
dlockhart: I'm surprised selecting an item in the menu doesn't automatically close it?

AdmiralCoco: I also expected the menu to close automatically - my guess is that there's something going on with the events passing through shadow-dom and not reaching the `d2l-sort-by`'s `d2l-menu`

I'll do some more testing and see if changing the `d2l-sort-by-option` to use `d2l-menu-item-radio` will have any effects

d2l-sort-by.html
```diff
@@ -0,0 +1,97 @@
...
+						type: String,
+						computed: 'getButtonText(selected)'
+					},
+					/**
+					* The currently selected option
+					*/
+					selected: {
+						type: String,
+						value: 'None', // TODO: langterm for default(none) selection
+						reflectToAttribute: true,
```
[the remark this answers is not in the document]
AdmiralCoco: I asked @dbatiste about this just now actually, and he had a similar explanation 🙂this was the main concern I had (i.e., just having the text) - it's clear now that I should use `d2l-menu-item-radio` here instead

AdmiralCoco:
I spoke with @cheics-d2l and he suggested naming this `d2l-sort-by-dropdown`, which would line up with your comment about not limiting ourselves in the future @dlockhart

d2l-sort-by.html
cheics-d2l: Selecting an item in the menu SHOULD automatically close it for sorting.

d2l-sort-by.html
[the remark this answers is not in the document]
cheics-d2l: > One open question I have that may need some collaboration with the design: what's the plan for sort order?

I'm working on this and getting this reviewed Friday. I'm leaning towards NOT having seperate entries for A-Z & Z-A, at least in a sort dropdown.

d2l-sort-by.html
```diff
@@ -0,0 +1,97 @@
...
+
+<dom-module id="d2l-sort-by">
+	<template strip-whitespace>
+		<style>
+			:host {
+				display: inline-block;
+			}
+		</style>
+		<d2l-dropdown-button-subtle text="[[_buttonText]]">
+			<d2l-dropdown-menu no-padding no-auto-fit id="d2l-sort-by-menu">
```
dlockhart: One thing I missed in the first pass: the `my-courses` screenshot I posted put the label in a visible position here above the menu. I'll include it again for context:

screen shot 2018-10-24 at 2 06 24 pm

It's the "Sort By" thing at the top. Probably a question for @cheics-d2l: since I'm assuming the end goal is to replace the my-courses usage with this component, do we want that same look here, or would we prefer to not have that visible label (there'd still be an ARIA label on the menu)?

cheics-d2l: Right now we're leaning towards removing the visible label on the Sort dropdown active state. We will likely have some sort of header for Filters though.

The objective is to replace all sorting orders with this component :)

d2l-sort-by.html
AdmiralCoco: I think figured out why this is happening: 

`d2l-sort-by-option` is just a `<d2l-menu-item-radio>`, and in the [code](https://github.com/BrightspaceUI/menu/blob/master/d2l-menu-item-radio.html#L56), it queries the parent (expecting a `d2l-menu`) for all the `d2l-menu-item-radio` elements and changes their `selected` state 

However, when that code runs in this context, the parent is `shadowRoot` (i.e., this element). It seems I can't just wrap the `d2l-menu-item-radio` as it is currently

dbatiste: I think the behavior for selectable item is intentional.  In the case of `d2l-menu-item-checkbox` elements, the expectation is that the user or consumer closes the dropdown because `d2l-dropdown-menu` expects that the user may make more than one change.  `d2l-menu-item-radio` shares the "selectable" menu item behavior, and it may be possible to optimize the close logic for that case.

For standard `d2l-menu-item` components, the [d2l-dropdown-menu](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-menu.html#L96) behavior listens for the `d2l-menu-item-select` event and closes the menu if the item is a `d2l-menu-item` because the expectation is that the user will typically only perform one action.

dbatiste: In terms of wrapping the menu items, I think that may not work well in a variety of ways, for a bunch of reasons.  The `d2l-menu` expects menu items to be inside.  If it's necessary to wrap them, then a custom menu item should be used, but I do not advise creating custom menu items.  Rather, I suggest using the built-in ones.

d2l-sort-by.html
dbatiste: If the idea of a header inside of dropdown content comes up more than once, I think it's something that should may be best handled by `d2l-dropdown` content components.  I've seen other version as well that has a tabby header thing.  I think we should aim for alignment on this.

d2l-sort-by-option.html
```diff
@@ -0,0 +1,46 @@
...
+			}
+		</style>
+		<d2l-menu-item text="[[text]]"></d2l-menu-item>
+	</template>
+	<script>
+		/**
+		 * `<d2l-sort-by-option>`
+		 * Option for sort by dropdown
+		 */
+		class SortByOption extends Polymer.mixinBehaviors(
```
dbatiste: Similar to my other comment in the "close" thread, it is very strange that this `SortByOption` component be a menu-item (i.e. by including MenuItemBehavior), which itself has a `d2l-menu-item` as a direct child.  `d2l-menu` should be the only direct parent of `role="menuitem"`.  This may break a11y and other menu behaviors.  The only time we should have menu items inside of menu items is for nested menus, where there is `d2l-menu` > `role="menuitem"` > `d2l-menu` > `role="menuitem"` ... etc.

d2l-sort-by.html
dbatiste: Note: `no-padding` should be the default for `d2l-dropdown-menu`, so it should be possible to omit this here.  (all our our menus should go edge-to-edge in dropdowns)

d2l-sort-by-option.html
AdmiralCoco: Yeah my attempted fix resulted in this basically becoming a custom menu item

The intent was to keep the elements named as they are, but have the same behaviour as the `d2l dropdown-menu`. It seems like it might cause more issues than it solves

Given this, would you/@dlockhart suggest changing the structure of this component to require a `d2l-dropdown` to be used?

dbatiste: Yep - that's certainly doable.  The majority of the logic is for menu items and selectable menu items is already in a behavior, so it would just be a little bit of radio-specific logic that gets moved.  So there would be a new behavior along these lines:
```
D2L.PolymerBehaviors.MenuItemRadioBehaviorImpl = {
	...   
};

D2L.PolymerBehaviors.MenuItemRadioBehavior = [
	D2L.PolymerBehaviors.MenuItemRadioBehaviorImpl,
	D2L.PolymerBehaviors.MenuItemSelectableBehavior
];
```
Then, in a similar way as this [custom menu-item demo](https://github.com/BrightspaceUI/menu/blob/master/demo/demo-components.html#L8), a sort-specific selectable menu item could be defined (using the new radio behavior).

AdmiralCoco: Sounds good, I'll work on creating the `MenuItemRadioBehavior`

d2l-sort-by-dropdown.html
```diff
@@ -0,0 +1,106 @@
...
+					* Label for the dropdown-menu
+					*/
+					label: {
+						type: String,
+						value: ''
+					},
+					/**
+					* The text of the currently selected option
+					*/
+					text: {
```
dbatiste: Should this be private?

d2l-sort-by-dropdown.html
```diff
@@ -0,0 +1,106 @@
...
+				this.text = e.target.text;
+				this.value = e.detail.value;
+
+				this.dispatchEvent(new CustomEvent(
+					'd2l-sort-by-dropdown-change',
+					{ bubbles: true, composed: true, detail: { value: this.value } }
+				));
+			}
+
+			getButtonText(selection) {
```
dbatiste: Nit: I'd probably name this something like `selectedOptionText` for clarity.

bower.json
```diff
@@ -16,7 +15,10 @@
     "wct.conf.json"
   ],
   "dependencies": {
-    "polymer": "Polymer/polymer#^2.0.0"
+    "polymer": "Polymer/polymer#^2.0.0",
+    "d2l-dropdown": "^6.6.2",
+    "d2l-menu": "^1.2.0",
+    "d2l-icons": "^5.1.2"
```
dbatiste: Nit: general best practice is to order the deps alphabetically.  Minor thing especially with a list this short... just calling it out since it will probably be raised elsewhere in the future so good to get in the habit.  :)

d2l-sort-by-dropdown-option.html
```diff
@@ -0,0 +1,44 @@
...
+			[
+				D2L.PolymerBehaviors.MenuItemRadioBehavior,
+			],
+			Polymer.Element
+		) {
+			static get is() { return 'd2l-sort-by-dropdown-option'; }
+			static get properties() {
+				return {
+					text: {
+						type: String,
```
dbatiste: Trailing commas.  Sorta surprised eslint did not catch.

AdmiralCoco: I don't think this rule is on for [`eslint-recommended`](https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js#L28)? 
(If I'm understanding how the config is resolved correctly)

AdmiralCoco:
Do you have any further comments for this PR @dbatiste, @dlockhart?

d2l-sort-by-dropdown.html
```diff
@@ -0,0 +1,105 @@
...
+					'd2l-sort-by-dropdown-change',
+					{ bubbles: true, composed: true, detail: { value: this.value } }
+				));
+			}
+
+			getSelectedOptionText(selection) {
+				// TODO: Add langterm with placeholder for localization
+				return selection
+					? `Sort by: ${selection}`
+					: 'Sort by';
```
dlockhart: If you don't want to wait for translations, you could likely steal them all from the LMS:
http://search.dev.d2l/source/xref/Lms/ep/lang/langterms/definitions/eP_MyContent.xml#81

AdmiralCoco: Thanks for the heads up! I'll add the localization in another PR 👍

PR #2, opened 2018-10-24
Initial implementation of d2l-search-facets

AdmiralCoco:
Updated usage:

`<d2l-search-facets>`: 
- Events:
    - Fires an event that indicates the grouping/value that changed, and whether the option is selected(checked) or not
        - e.g., `{ checked: true, option: "online", grouping: "format" }`

`<d2l-search-facets-grouping>`:
- Attributes:
    - `text` (label of the grouping)
    - `value`
    - `hide-after` (how many options are shown initially, default is all)
- Events:
    - Adds grouping to the event to the event fired by`<d2l-search-facets-option>`

`<d2l-search-facets-option>`:
- Attributes:
    - `text` (of the option)
    - `value`
    - `count` (optional)
- Events:
    - Fires an event on selection
        - e.g., `{ option: …, checked: true/false }`

---
screen shot 2018-10-24 at 11 57 06 am

Example usage:
```html
<d2l-search-facets>
  <d2l-search-facets-grouping value="status" text="My Status">
    <d2l-search-facets-option value="new-courses" text="New Courses Only" count="21"></d2l-search-facets-option>
  </d2l-search-facets-grouping>
  <d2l-search-facets-grouping value="format" text="Format" hide-after="1">
    <d2l-search-facets-option value="online" text="Online"></d2l-search-facets-option>
    <d2l-search-facets-option value="in-class" text="In-Class"></d2l-search-facets-option>
  </d2l-search-facets-grouping>
</d2l-search-facets>
```

### DIFF
--- a/d2l-search-facets-grouping.html
+++ b/d2l-search-facets-grouping.html
@@ -137,25 +137,12 @@
 					this._boundListeners._onFacetOptionChange);
 			}
 
-			_getEffectiveChildren() {
-				return Polymer.FlattenedNodesObserver.getFlattenedNodes(this)
-					.filter(function(n) { return n.nodeType === Node.ELEMENT_NODE; });
-			}
-
 			_onFacetOptionChange(e) {
-				e.detail.checked
-					? this._optionsSelected.push(e.detail.value)
-					: this._optionsSelected.splice(this._optionsSelected.indexOf(e.detail.value), 1);
-
-				this.dispatchEvent(new CustomEvent(
-					'd2l-search-facets-grouping-change', {
-						bubbles: true, composed: true, detail:
-						{
-							grouping: this.value,
-							selected: this._optionsSelected,
-						}
-					}
-				));
+				this.dispatchEvent(new CustomEvent('d2l-search-facets-grouping-change', {
+					bubbles: true,
+					composed: true,
+					detail: Object.assign({}, e.detail, { grouping: this.value })
+				}));
 			}
 
 			_hideOptions() {
--- a/d2l-search-facets-grouping.html
+++ b/d2l-search-facets-grouping.html
@@ -49,6 +49,17 @@
 			static get properties() {
 				return {
 					/**
+					* Reference to bound listener functions to be removed in disconnectedCallback
+					*/
+					_boundListeners: {
+						type: Object,
+						value: function() {
+							return {
+								_onFacetOptionChange: null,
+							};
+						}
+					},
+					/**
 					* List of <d2l-search-facets-option> elements
 					*/
 					_optionElements: {
@@ -70,62 +81,83 @@
 						value: false,
 					},
 					/**
-					* Label of the grouping
+					* Number of facet options initially displayed. All options are shown by default
 					*/
-					label: {
+					hideAfter: {
+						type: Number,
+					},
+					/**
+					* The name of the grouping
+					*/
+					text: {
 						type: String,
 						value: '',
 					},
 					/**
-					* Number of facet options initially displayed. All options are shown by default
+					* The value of the grouping name
 					*/
-					initiallyShown: {
-						type: Number,
-					}
+					value: {
+						type: String,
+						value: '',
+					},
 				};
 			}
 
 			connectedCallback() {
 				super.connectedCallback();
-				this.addEventListener('d2l-search-facets-option-change', function(e) {
-					e.detail.checked
-						? this._optionsSelected.push(e.detail.option)
-						: this._optionsSelected.splice(this._optionsSelected.indexOf(e.detail.option), 1);
-
-					this.dispatchEvent(new CustomEvent(
-						'd2l-search-facets-grouping-change', {
-							bubbles: true, composed: true, detail:
-							{
-								grouping: this.label,
-								selected: this._optionsSelected,
-							}
-						}
-					));
-				}.bind(this));
-
-				// Child elements may not be rendered if this is a <dom-repeat>. Wait for next
-				// render to query for them
 				Polymer.RenderStatus.afterNextRender(this, function() {
-					this._optionElements = Polymer.FlattenedNodesObserver.getFlattenedNodes(this)
-						.filter(function(n) { return n.nodeType === Node.ELEMENT_NODE; });
+					// Child elements may not be rendered if this is a <dom-repeat>. Wait for next
+					// render to query for them
+					this._boundListeners._onFacetOptionChange = this._onFacetOptionChange.bind(this);
 
-					if (!this.initiallyShown) {
-						this.initiallyShown = this._optionElements.length;
+					this.addEventListener('d2l-search-facets-option-change',
+						this._boundListeners._onFacetOptionChange);
+
+					if (!this.hideAfter) {
+						this.hideAfter = this._getOptionElements().length;
 					}
 
-					this._showMoreButtonVisible = this.initiallyShown < this._optionElements.length;
+					this._showMoreButtonVisible = this.hideAfter < this._getOptionElements().length;
 					this._updateOptionVisibility('none');
 				}.bind(this));
 			}
 
+			disconnectedCallback() {
+				super.disconnectedCallback();
+				this.removeEventListener('d2l-search-facets-option-change',
+					this._boundListeners._onFacetOptionChange);
+			}
+
+			_getOptionElements() {
+				return Polymer.FlattenedNodesObserver.getFlattenedNodes(this)
+					.filter(function(n) { return n.nodeType === Node.ELEMENT_NODE; });
+			}
+
+			_onFacetOptionChange(e) {
+				e.detail.checked
+					? this._optionsSelected.push(e.detail.value)
+					: this._optionsSelected.splice(this._optionsSelected.indexOf(e.detail.value), 1);
+
+				this.dispatchEvent(new CustomEvent(
+					'd2l-search-facets-grouping-change', {
+						bubbles: true, composed: true, detail:
+						{
+							grouping: this.value,
+							selected: this._optionsSelected,
+						}
+					}
+				));
+			}
+
 			_showMore() {
-				this._updateOptionVisibility('inline-block');
+				this._updateOptionVisibility('block');
 				this._showMoreButtonVisible = false;
 			}
 
 			_updateOptionVisibility(display) {
-				for (let index = this.initiallyShown; index < this._optionElements.length; index++) {
-					this._optionElements[index].style.display = display;
+				const optionElements = this._getOptionElements();
+				for (let index = this.hideAfter; index < optionElements.length; index++) {
+					optionElements[index].style.display = display;
 				}
 			}
 
--- a/d2l-search-facets-grouping.html
+++ b/d2l-search-facets-grouping.html
@@ -42,9 +42,9 @@
 		<fieldset class="d2l-search-facets-grouping-fieldset">
 			<legend class="d2l-search-facets-grouping-legend">[[text]]</legend>
 			<slot></slot>
-			<template is="dom-if" if="[[_showMoreButtonVisible]]">
+			<template is="dom-if" if="[[_showOptionsButtonVisible]]">
 				<d2l-input-checkbox-spacer class="d2l-search-facets-grouping-show-more">
-					<button on-click="_showMore">More</button>
+					<button on-click="_showOptions">More</button>
 				</d2l-input-checkbox-spacer>
 			</template>
 		</fieldset>
@@ -70,11 +70,11 @@
 						}
 					},
 					/**
-					* List of <d2l-search-facets-option> elements
+					* Reference to the style element created to hide options. Deleted when
+					* the show options button is clicked
 					*/
-					_optionElements: {
-						type: Array,
-						value: function() { return []; },
+					_hideOptionsStyleElement: {
+						type: HTMLElement,
 					},
 					/**
 					* List of selected options (e.g., ['optionA', ...]s)
@@ -84,14 +84,15 @@
 						value: function() { return []; },
 					},
 					/**
-					* Visibility of 'More' button - true if initiallyShown < _optionElements.length
+					* Visibility of 'More' button, true if hideAfter is present
 					*/
-					_showMoreButtonVisible: {
+					_showOptionsButtonVisible: {
 						type: Boolean,
 						value: false,
 					},
 					/**
-					* Number of facet options initially displayed. All options are shown by default
+					* Number of facet options initially displayed. All options are shown by default.
+					* Should not be greater than the number of child elements
 					*/
 					hideAfter: {
 						type: Number,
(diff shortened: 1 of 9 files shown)


AdmiralCoco:
@cheics-d2l suggested renaming this to `d2l-expandable-checkbox-group`, and to call the `initially-shown` attribute `hide-after` instead

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+					* Visibility of 'More' button - true if initiallyShown < _optionElements.length
+					*/
+					_showMoreButtonVisible: {
+						type: Boolean,
+						value: false,
+					},
+					/**
+					* Label of the grouping
+					*/
+					label: {
```
dlockhart: Normally when text is visible we name the attribute `text` and use `label` for either invisible labels or labels for form inputs. I think `text` would make the most sense here.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+						type: String,
+						value: '',
+					},
+					/**
+					* Number of facet options initially displayed. All options are shown by default
+					*/
+					initiallyShown: {
```
dlockhart: Reading the attribute name, I would have assumed it was a `boolean`. I might suggest calling this `num-shown` or `num-initially-shown`? I dunno.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+				this.addEventListener('d2l-search-facets-option-change', function(e) {
+					e.detail.checked
+						? this._optionsSelected.push(e.detail.option)
+						: this._optionsSelected.splice(this._optionsSelected.indexOf(e.detail.option), 1);
+
+					this.dispatchEvent(new CustomEvent(
+						'd2l-search-facets-grouping-change', {
+							bubbles: true, composed: true, detail:
+							{
+								grouping: this.label,
```
dlockhart: The label (or text) will be a localized language term, so exposing it to consumers won't be super helpful for them. If you think knowing the group will be important, I'd suggest adding a `value` property that developers can use to set a non-localized value.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+				font-weight: bold;
+			}
+
+		</style>
+		<span class="d2l-search-facets-grouping-label">[[label]]</span>
+		<d2l-input-checkbox-spacer></d2l-input-checkbox-spacer>
+		<slot></slot>
+		<template is="dom-if" if="{{_showMoreButtonVisible}}">
+			<d2l-input-checkbox-spacer class="d2l-search-facets-grouping-show-more">
+				<button on-click="_showMore">More</button>
```
dlockhart: I'm not sure what the design calls for in terms of what this button should look like... but I think we should be using one of our existing button styles instead of inventing something new. `<d2l-button-subtle>` seems like potentially a good choice? Maybe using the `<d2l-more-less>` "more" icon?

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			.d2l-search-facets-grouping-show-more > button:hover,
+			.d2l-search-facets-grouping-show-more > button:focus {
+				cursor: pointer;
+				font-weight: bold;
+			}
+
+		</style>
+		<span class="d2l-search-facets-grouping-label">[[label]]</span>
```
dlockhart: Seems like a good opportunity to use `<fieldset>` and `<legend>` to group the checkboxes semantically together. To a screen reader, it makes things super helpful, as when they visit a checkbox it will announce both the checkbox label (e.g. "New Courses only") but also the group legend text (e.g. "My Status").

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+						}
+					));
+				}.bind(this));
+
+				// Child elements may not be rendered if this is a <dom-repeat>. Wait for next
+				// render to query for them
+				Polymer.RenderStatus.afterNextRender(this, function() {
+					this._optionElements = Polymer.FlattenedNodesObserver.getFlattenedNodes(this)
+						.filter(function(n) { return n.nodeType === Node.ELEMENT_NODE; });
+
```
dlockhart: I wonder if you could use the `nth-child` CSS selector to do some really clever things here and avoid all this code. Specifically, I'm thinking of one of [the recipes](https://css-tricks.com/useful-nth-child-recipies/) that goes something like:

```css
// select all but the first 5
:nth-child(n+6) {
    display: none;
}
:host([all-visible]) :nth-child(n+6) {
    display: inline-block;
}
```

Browser support for that selector would be something to check on, and since the number itself is dynamic you may have to use JavaScript to add the CSS. But it could be cleaner to rely solely on CSS instead of needing to find all the children and update the visibility again later via JS.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+		</style>
+		<span class="d2l-search-facets-grouping-label">[[label]]</span>
+		<d2l-input-checkbox-spacer></d2l-input-checkbox-spacer>
+		<slot></slot>
+		<template is="dom-if" if="{{_showMoreButtonVisible}}">
+			<d2l-input-checkbox-spacer class="d2l-search-facets-grouping-show-more">
+				<button on-click="_showMore">More</button>
+			</d2l-input-checkbox-spacer>
```
dlockhart: TODO: localize the "More" text.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+			.d2l-search-facets-grouping-show-more > button:focus {
+				cursor: pointer;
+				font-weight: bold;
+			}
+
+		</style>
+		<span class="d2l-search-facets-grouping-label">[[label]]</span>
+		<d2l-input-checkbox-spacer></d2l-input-checkbox-spacer>
+		<slot></slot>
+		<template is="dom-if" if="{{_showMoreButtonVisible}}">
```
dbatiste: I think this could use one-way data-binding ... i.e. `if="[[...]]"`.  It would be good to get in a habit of not using two-way binding anywhere.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
+<link rel="import" href="../polymer/polymer-element.html">
+<link rel="import" href="../d2l-inputs/d2l-input-checkbox-spacer.html">
+<link rel="import" href="../d2l-typography/d2l-typography-shared-styles.html">
+
+<dom-module id="d2l-search-facets-grouping">
+	<template strip-whitespace>
+		<style>
+			:host {
+				display: flex;
```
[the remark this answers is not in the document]
dlockhart: If the parent is a `block` element, I actually don't think you need to use flex layout here at all. If these were also block elements, they'd naturally stack on top of each other.

d2l-search-facets-grouping.html
dlockhart: More/less may not be totally appropriate since it's all about expanding/collapsing a bunch of content, whereas you just want to hide it initially and then expand it forever I think? But under the covers it uses a subtle button, which might work well for you.

d2l-search-facets-grouping.html
dlockhart: Yeah you'd need to dynamically add the rule via JavaScript (and remove it and re-add if the attribute value changes). I don't have a ton of experience doing this so you may need to experiment!

d2l-search-facets-option.html
```diff
@@ -0,0 +1,97 @@
...
+				@apply --d2l-body-compact-text;
+			}
+		</style>
+		<d2l-input-checkbox
+			checked={{checked::change}}
+			disabled=[[disabled]]
+			name="[[name]]"
+			on-change="_handleChange"
+			class="d2l-search-facets-option-checkbox"
+		>[[_facetText]]</d2l-input-checkbox>
```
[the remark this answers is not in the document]
dlockhart: Ah, right. What you have is good then!

d2l-search-facets-grouping.html
cheics-d2l: I can confirm - Hide initially, expand _forever_ is the desired behaviour for this control as part of the "Facetted Search Panel" pattern.

The "expand forever" behaviour may be desired by several other components as part of the MORE/LESS pattern, and might be worth doing consistently in the MORE/LESS control.

(Note the example of the GitHub "5 Invisible Conversations" & "Show" pattern on this page)

d2l-search-facets-option.html
```diff
@@ -0,0 +1,97 @@
...
+					*/
+					checked: {
+						type: Boolean,
+						reflectToAttribute: true,
+					},
+					/**
+					* The number of search results for this option
+					*/
+					count: {
+						type: String,
```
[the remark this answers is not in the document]
dlockhart: That could be an interesting feature that the component does on its own... like you could set a threshold after which it would say `10+` or `100+`? Not sure about this, using a string and losing the strongly-typed data is unfortunate.

d2l-search-facets-grouping.html
dlockhart: Oh neat, yeah that could be a cool addition to more/less. The other potentially limiting factor for more/less is that today the only way to set a threshold is via pixel height -- e.g. "clip this at a height of 200px". So only showing 3 or 5 or 10 children isn't an option currently. But it could be added to more/less probably easily-ish.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+								selected: this._optionsSelected,
+							}
+						}
+					));
+				}.bind(this));
+
+				// Child elements may not be rendered if this is a <dom-repeat>. Wait for next
+				// render to query for them
+				Polymer.RenderStatus.afterNextRender(this, function() {
+					this._optionElements = Polymer.FlattenedNodesObserver.getFlattenedNodes(this)
```
dbatiste: It looks like we end up with 2 sources of truth here - the DOM, and this property.  If additional options are added or removed after this code runs, they would not be reflected in this array.  In addition, if options are removed, this component would still be holding references to those DOM nodes.  I think it is better to treat the DOM as the single source of truth.  But I could be misunderstanding how this is used as well... ?

d2l-search-facets-option.html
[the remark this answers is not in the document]
cheics-d2l: Generally for facetted search, the UI just shows exact numbers:
image
[Library Search Example](https://primo.tug-libraries.on.ca/primo_library/libweb/action/search.do?fn=search&ct=search&initialSearch=true&mode=Basic&tab=search&indx=1&dum=true&srt=rank&vid=WATERLOO&frbg=&tb=t&vl%28freeText0%29=electricity&scp.scps=scope%3A%28%22CENTRAL%22%29%2Cscope%3A%28GUELPH%29%2Cscope%3A%28LAURIER%29%2Cscope%3A%28WATERLOO%29%2Cprimo_central_multiple_fe)

I can see possible use cases where DEV constraints might need to return an approximate number, but not sure if there's cases where Design would specify 10+ or the like. I would lean towards using number and loosening the constraint in the future IF we need it.

From other design constraint standpoint, if the number is 5,012,123,123  ( (5B+) (??) ) -- we might not want to show the entire number since it could get sorta ugly with weird wrapping. But NOT showing the number could also be confusing in a selection scenario. Again, can't imagine what data would result in 5 Billion, so I would lean towards number only.

d2l-search-facets-grouping.html
```diff
@@ -0,0 +1,135 @@
...
+					*/
+					initiallyShown: {
+						type: Number,
+					}
+				};
+			}
+
+			connectedCallback() {
+				super.connectedCallback();
+				this.addEventListener('d2l-search-facets-option-change', function(e) {
```
dbatiste: It looks like this event handler is added but never removed.  If this element were attached, detached, and re-attached to the DOM, 2 event handlers would be wired up so this logic would be called twice.  Best practice when wiring up event-handlers in `connectedCallback` is to also un-wire them in `disconnectedCallback`.

It may be possible to defer this work to `afterNextRender`.  I would typically do this unless there is some reason not to.

d2l-search-facets.html
```diff
@@ -0,0 +1,58 @@
...
+					optionsSelected: {
+						type: Object,
+						value: function() { return {}; },
+					}
+				};
+			}
+
+			connectedCallback() {
+				super.connectedCallback();
+				this.addEventListener('d2l-search-facets-grouping-change', function(e) {
```
dbatiste: Ditto here for the event handler.

d2l-search-facets-grouping.html
dlockhart: That's what I figured you'd need. If you keep a reference to the `<style>` element, can you then just remove it when they click the "more" button"? You could also remove and re-add it if the `hide-after` attribute value changes.

dlockhart: Also, since this is Polymer 2 and will get transpiled, you can use the backtick strings instead of needing to concatenate 💥

AdmiralCoco: Yeah, that's what I was thinking of doing 👍 
Cool, I'll definitely use template strings instead then

AdmiralCoco: Just wondering, did you have a use case in mind where the `hide-after` attribute would change after it's initially set?

cheics-d2l: Personally - I remember when we were chatting about this at the whiteboard.

The 2 scenarios for "Hide after":
- (1) There's a TON of facet options (like 500 departments) and showing them all would be a ton of work
- (2) There's a few facet options, but only a couple have actual values

Ex2:
- [ ] English (10)
- [ ] French (2)
- MORE
- [ ] Arabic (0)
- [ ] German (0)
- Etc.

Seems like we're covering situation 1 very well with the hide after, but not situation 2. Based on the situations, I'm not sure if there's times where we would need to dynamically update the `hide-after`... unless we're implementing the behaviour of situation 2 through the `hide-after` property and back end stuff.

dbatiste: I think adding a style block dynamically will cause a reflow similar to the original code.   Technically the new style block should be scoped as well, and I suspect shadyCSS would not scope this without some extra call to shadyCSS.  I'm not sure if this gets us anywhere.

AdmiralCoco: @cheics-d2l I think in situation 2 where the consumer could know those numbers up front, it can be set up (on the consumer side) so that those options wouldn't be in the list to be rendered

@dbatiste Not too familiar with shadyCSS but I'm assuming your concern is that adding it in this way to the shadowRoot isn't enough to keep it scoped to this element?

dlockhart: Also... if there are like 500 departments that would show up, maybe we shouldn't be rendering them at all? That's a lot of HTML to send down and DOM to create if it's just going to be hidden initially. In this case, I'd want us to lazily load them when the user clicks "more"... but now that's going beyond what this web component can do, since it doesn't know anything about how to get those items.

AdmiralCoco: Regarding the handling of having many options - perhaps we could have an attribute (e.g., `remote-source`) that can be set on the grouping? (And `hide-after` wouldn't be set)

In this case, the code for adding/removing of CSS won't be run, but the 'More' button will be shown. Clicking the button will fire an event:
  - (`d2l-search-facet-grouping-more, { grouping: ...}`)

to tell the consumer to fetch more options

It will be up to the consumer to then add those elements (and also update the visibility of the 'More' button)

dlockhart: I really hesitate to bake in too much logic like remote sources... at that point this simple component seems like it's doing too much.

What if we make it even simpler, where the group just has a single `bool` attribute called `has-more`. If that attribute is present, the more button is displayed and clicking it fires the event. It's then on the consumer to fetch more items and insert them as children of the group, and remove the `has-more` attribute if there aren't more.

AdmiralCoco: Yeah that's exactly what I meant 🙂 `remote-source` was the first thing that came to mind (as I used it elsewhere) - `has-more` sounds good to me

dlockhart: Cool! Then really we can just remove all the auto hiding and `hide-after` stuff altogether I think, since the consumer can just put the first N or whatever in the group and set the `has-more` attribute and be done with it.

AdmiralCoco:
Thanks again for the feedback 🙂

PR #3, opened 2018-10-29
Feature/search results count

MichaelKovalchuk:


bower.json
```diff
@@ -15,7 +15,8 @@
     "wct.conf.json"
   ],
   "dependencies": {
-    "polymer": "Polymer/polymer#^2.0.0"
+    "polymer": "Polymer/polymer#^2.0.0",
+    "d2l-localize-behavior": "BrightspaceUI/localize-behavior#^1.1.3"
```
AdmiralCoco: Nit: reorder dependencies alphabetically

d2l-search-results-count.html
```diff
@@ -0,0 +1,72 @@
+<link rel="import" href="../polymer/polymer-element.html">
+<link rel="import" href="d2l-search-results-count-localize-behavior.html">
+
+<dom-module id="d2l-search-results-count">
+	<template strip-whitespace>
+		<style>
```
AdmiralCoco: You may need to add an empty rule due to linting issues ([example](https://github.com/BrightspaceUI/menu/blob/master/d2l-menu-item-radio.html#L16))

test/d2l-search-results-count.html
```diff
@@ -0,0 +1,61 @@
...
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+		<title>d2l-search-results-count test</title>
+		<script src="../../webcomponentsjs/webcomponents-loader.js"></script>
+		<script src="../../web-component-tester/browser.js"></script>
+		<script src="../../iron-test-helpers/mock-interactions.js"></script>
+		<script src="../../iron-test-helpers/test-helpers.js"></script>
+		<link rel="import" href="../all-imports.html">
```
AdmiralCoco: Just to keep the import consistent with the [demo file](https://github.com/BrightspaceUI/facet-filter-sort/pull/3/files#diff-8530dd38feb4e1a55635714f592cf176R11)

```suggestion
		<link rel="import" href="../d2l-search-results-count.html">
```

Or change both to all-imports

test/d2l-search-results-count.html
```diff
@@ -0,0 +1,61 @@
...
+			</template>
+		</test-fixture>
+		<script>
+			describe('<d2l-search-results-count>', function() {
+				describe('basic', function() {
+					let basicFixture;
+
+					beforeEach(function(done) {
+						basicFixture = fixture('basic');
+						Polymer.RenderStatus.afterNextRender(basicFixture, function() {
```
AdmiralCoco: You can pass in `done` if you don't need to do any other initialization:
```suggestion
						Polymer.RenderStatus.afterNextRender(basicFixture, done);
```

test/d2l-search-results-count.html
```diff
@@ -0,0 +1,61 @@
...
+					beforeEach(function(done) {
+						basicFixture = fixture('basic-no-query');
+						Polymer.RenderStatus.afterNextRender(basicFixture, function() {
+							done();
+						});
+					});
+
+					it('text should match', function() {
+						const shadowRoot = basicFixture.shadowRoot;
+						expect(shadowRoot.textContent).to.contain('1 - 20 of 22 results');
```
AdmiralCoco: Is there any particular reason why `contain` is used vs. `equal`?

d2l-search-results-count.html
```diff
@@ -0,0 +1,72 @@
...
+					},
+					/**
+					* Range start
+					*/
+					rangeStart: {
+						type: Number,
+						value: 0
+					},
+					/**
+					* Range to
```
AdmiralCoco: Minor: should be 'Range end'

test/d2l-search-results-count.html
```diff
@@ -0,0 +1,97 @@
...
+					let basicFixture;
+
+					beforeEach(function(done) {
+						basicFixture = fixture('paged');
+						Polymer.RenderStatus.afterNextRender(basicFixture, done);
+					});
+
+					it('text should match', function() {
+						const shadowRoot = basicFixture.shadowRoot;
+						const text = shadowRoot.textContent.replace(/[\n\r]+|[\s]{2,}/g, ' ').trim();
```
AdmiralCoco: What does this regex match for?

MichaelKovalchuk: replaces tabs and new lines with spaces. If followed by trim() - removes the whitespace and leaves the content text.

AdmiralCoco: Interesting - in the past I've used `innerHTML` and it didn't require any regex; I'm not sure if it'll make a difference though

MichaelKovalchuk: i guess I could use innerHtml, but for some reason it wouldn't return the values. textContent works much faster, but we are loosing the speed advantage at parsing and trimming.

PR #16, opened 2019-02-15
Use tabs instead

svanherk:
Let's try this Draft PR thing instead.

Did a bit of demo cleanup, it wasn't working in Edge and the typography wasn't being set.

I found a couple other issues around accessibility I was gonna tackle in another PR, so @CarolBusman you can keep focusing on the wrapper component.

~**WIP**~
~Still working on test fixes~
~Still need to cleanup some unused functions~

### DIFF
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown.js
@@ -18,7 +18,7 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 	static get template() {
 		return html`
 			<style include="d2l-filter-dropdown-styles"></style>
-			<d2l-dropdown-button-subtle text="[[_numFiltersText(_numFiltersSelectedTotal)]]">
+			<d2l-dropdown-button-subtle text="[[_selectedFilterCountText(_selectedFilterCount)]]">
 				<d2l-dropdown-content
 					min-width="[[minWidth]]"
 					max-width="[[maxWidth]]"
@@ -73,9 +73,9 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 				type: String,
 				value: 'd2l-filter-dropdown-closed'
 			},
-			_numFiltersSelectedTotal: {
+			_selectedFilterCount: {
 				type: Number,
-				computed: '_getTotalNumFiltersSelected(_filters.*)'
+				computed: '_getSelectedFilterCount(_filters.*)'
 			}
 		};
 	}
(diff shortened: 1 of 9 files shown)


svanherk:
@CarolBusman I still need to fix the tests, but I think the main components and styles are stable now and ready for any feedback.  I did adjust some names you may be relying on, but I think the changes should be minor

components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -19,41 +18,25 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
...
-					no-padding
-					render-content
-					no-pointer>
-					<div class="d2l-filter-dropdown-content-header">
-						<span>[[localize('filterBy')]]</span>
-						<button hidden$="[[!_numFilters]]" class="d2l-filter-dropdown-clear-button" on-click="_clearFilters">[[localize('clear')]]</button>
-					</div>
-					<div class="d2l-filter-dropdown-tab-selector">
-						<dom-repeat items="[[_filters]]">
+			<d2l-dropdown>
```
dbatiste: One too many layers here - we should be able to remove the `d2l-dropdown` (generic opener) since `d2l-dropdown-button-subtle` is an opener.  The generic `d2l-dropdown` opener is only used when we want to define our own opener element and denote it with `class="d2l-dropdown-opener"`.  This is an awkward aspect of the component names, and the fact that so many things are possible.

components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -81,67 +71,40 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
...
-			_selectedTab: {
-				type: String,
-				computed: '_getSelectedTab(_filters.*)'
-			},
 			_filterDropdownClosed: {
 				type: String,
 				value: 'd2l-filter-dropdown-closed'
 			},
-			_numFilters: {
+			_numFiltersSelectedTotal: {
```
dbatiste: Super nit: I might be inclined to name this `_selectedFilterCount`.  Meh.

components/d2l-filter-dropdown/d2l-filter-dropdown.js
svanherk: Thanks Dave!  Totally did not realize that, I agree that the naming isn't the most intuitive, but I also didn't read the README close enough

components/d2l-filter-dropdown/d2l-filter-dropdown.js
svanherk: I wasn't particularly happy with any of my re-naming lol, so sure!

components/d2l-filter-dropdown/d2l-filter-dropdown.js
dbatiste: Maybe in hindsight we should have named the generic opener `d2l-dropdown-generic`?  Then it might be more obvious that it's not meant to be a wrapper around the other prebuilt openers.

svanherk:
Ok, this is ready.  @CarolBusman if you're fine with this, it's probably easiest to just merge this into your branch and then Dave and Dawing only have to review https://github.com/BrightspaceUI/facet-filter-sort/pull/14

PR #19, opened 2019-02-27
Fixing typo in event dispatch.

CarolBusman:


components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -235,7 +235,7 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 	_dispatchFilterSelectionChanged(key) {
 		this.dispatchEvent(
 			new CustomEvent(
-				this._filterSelectionChanged,
+				this._filterSelectedChanged,
```
dbatiste: What's the rationale for having this private property?  Could this just be `d2l-filter-selected-changed`?

CarolBusman: I put the event name strings in the properties so that they'd all be grouped together near the top of the class. Should we just be using the string directly here?

dbatiste: If all we need is a string, then I would just use that and forego the property which has some extra baggage that comes with it.

PR #20, opened 2019-03-01
Adding option changed event on clear.

CarolBusman:


svanherk:
Would it be easier for you to get a new type event?  `d2l-filter-dropdown-cleared`?

PR #23, opened 2019-03-08
Cbusman/add option update function

CarolBusman:


components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -138,6 +138,17 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 		}
 	}
 
+	updateFilterOptionSelectedStatus(categoryKey, optionKey, value) {
+		const cIndex = this._getFilterIndexFromKey(categoryKey);
+		if (cIndex >= 0) {
+			const oIndex = this._getOptionIndexFromKey(cIndex, optionKey);
+			if (oIndex >= 0 && this._filters[cIndex].options[oIndex].selected !== value) {
```
jeroach: I assume that the checks you've done ensure that `selected` and `numSelected` on the lower line can never be null?

If that's true, then cool :)

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -47,6 +47,18 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
 		return expectedNum;
 	}
 
+	function _updateExpectedOption(index, value, expected) {
+		for (let i = 0; i < expected.length; i++) {
+			const cat = expected[i];
+			for (let j = 0; j < cat.options.length; j++) {
+				if (options[index].cat === cat.key && options[index].key === cat.options[j].key && cat.options[j].selected !== value) {
```
jeroach: Where is `options` defined? Am I blind ;)

CarolBusman: Options is defined [here](https://github.com/BrightspaceUI/facet-filter-sort/blob/cbusman/add_option_update_function/test/d2l-filter-dropdown/d2l-filter-dropdown.js#L8) and populated [here](https://github.com/BrightspaceUI/facet-filter-sort/blob/cbusman/add_option_update_function/test/d2l-filter-dropdown/d2l-filter-dropdown.js#L87).

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -271,6 +283,33 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
 				done();
 			});
 		});
+		test('updating filter option updates filters', function() {
+			const update = 2;
+			_setSelectedOptions([0, 4, 5]);
```
jeroach: Is there some significance to what these numbers mean? If so, would it be more clear in an enum?

Or if they're just the place in the list then ignore me :)

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -271,6 +283,33 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
...
+			filter.updateFilterOptionSelectedStatus(options[update].cat, options[update].key, true);
+			_updateExpectedOption(update, true, expected);
+			assert.deepEqual(expected, filter._filters);
+		});
+		test('updating filter option updates display', function(done) {
+			const update = 2;
+			const selected = [0, 4, 5];
+			_setSelectedOptions(selected);
+			const expected = _getExpectedAndImport(filter);
+			filter.updateFilterOptionSelectedStatus(options[update].cat, options[update].key, true);
```
jeroach: Again, I don't see where `options` is defined. If it's a file level private variable, should we prefix it with an underscore? e.g. `_options`

test/d2l-filter-dropdown/d2l-filter-dropdown.js
CarolBusman: Yup; they're just indices for the options that we want to select for the test. There's nothing notable about them in specific, other than the fact that they cause the test to end up with different amounts of selected options per category.

test/d2l-filter-dropdown/d2l-filter-dropdown.js
jeroach: Never mind this - was addressed above, and I don't think that we need the underscore prefix

PR #25, opened 2019-03-20
Matching strings toLower instead of using regex to avoid having to escape special characters

CarolBusman:


components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
```diff
@@ -99,13 +99,18 @@ class D2LFilterDropdownPage extends mixinBehaviors([D2L.PolymerBehaviors.FilterD
...
 		if (clear || value.length) {
-			var regex = new RegExp(value, 'i');
 			for (var i = 0; i < this.options.length; i++) {
-				this._setOptionProp('display', clear || regex.test(this.options[i].title), i);
+				this._setOptionProp('display', clear || this._caseInsensitiveContainsSubstring(this.options[i].title, value), i);
 			}
 		}
 	}
 
+	_caseInsensitiveContainsSubstring(s, sub) {
```
jeroach: Would it be possible to add a few unit tests for this method?

PR #27, opened 2019-03-27
Emit search field changed event instead of searching here

AlexBedley:
This PR moves searching out of this component. Instead this component will emit an event, allowing parent components to handle search however they choose.

This is the first step to allow us to search server-side and it also unblocks pagination for filters.

Paired with: https://github.com/BrightspaceHypermediaComponents/common/pull/17

### DIFF
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
@@ -22,7 +22,7 @@ class D2LFilterDropdownPage extends mixinBehaviors([D2L.PolymerBehaviors.FilterD
 			<d2l-menu label="[[parentTitle]]">
 				<dom-repeat items="[[options]]" as="o">
 					<template>
-					<d2l-menu-item-checkbox text="[[o.title]]" value="[[o.key]]" hidden$="[[!o.display]]" selected=[[o.selected]]></d2l-menu-item-checkbox>
+					<d2l-menu-item-checkbox text="[[o.title]]" value="[[o.key]]" selected=[[o.selected]]></d2l-menu-item-checkbox>
 					</template>
 				</dom-repeat>
 			</d2l-menu>
@@ -45,8 +45,7 @@ class D2LFilterDropdownPage extends mixinBehaviors([D2L.PolymerBehaviors.FilterD
 					// {
 					// 	key: '',
 					// 	title: '',
-					// 	selected: false,
-					// 	display: true
+					// 	selected: false
 					// }
 				]
 			},
@@ -96,19 +95,19 @@ class D2LFilterDropdownPage extends mixinBehaviors([D2L.PolymerBehaviors.FilterD
 	}
 
 	_handleSearchChange(e) {
-		var value = e.detail.value;
-		var clear = value === '';
-		if (clear || value.length) {
-			for (var i = 0; i < this.options.length; i++) {
-				this._setOptionProp('display', clear || this._caseInsensitiveContainsSubstring(this.options[i].title, value), i);
-			}
-		}
-	}
-
-	_caseInsensitiveContainsSubstring(s, sub) {
-		const lowerS = s.toLowerCase();
-		const lowerSub = sub.toLowerCase();
-		return lowerS.indexOf(lowerSub) >= 0;
+		this.dispatchEvent(
+			new CustomEvent(
+				'd2l-filter-dropdown-page-searched',
+				{
+					detail: {
+						value: e.detail.value,
+						categoryKey: this.parentKey
+					},
+					composed: true,
+					bubbles: true
+				}
+			)
+		);
 	}
 
 	_handleMenuItemChange(e) {
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown.js
@@ -64,7 +64,7 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 					// 	options: [{
 					// 		key: '',
 					// 		title: '',
-					// 		selected: false,
+					// 		selected: false
 					// 	}]
 					// }
 				]
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown.js
@@ -156,6 +156,18 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 		}
 	}
 
+	setFilterOptions(categoryKey, options) {
+		const index = this._getFilterIndexFromKey(categoryKey);
+		if (index >= 0) {
+			this._setProp('options', [], index);
+			options.forEach(o => {
+				this.addFilterOption(categoryKey, o.key, o.title, o.selected);
+			});
+		} else {
+			throw new Error(`Cannot find category with key ${categoryKey}`);
+		}
+	}
+
 	updateFilterOptionSelectedStatus(categoryKey, optionKey, value) {
 		const cIndex = this._getCategoryIndexFromKey(categoryKey);
 		if (cIndex >= 0) {
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown.js
@@ -157,7 +157,7 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 	}
 
 	setFilterOptions(categoryKey, options) {
-		const index = this._getFilterIndexFromKey(categoryKey);
+		const index = this._getCategoryIndexFromKey(categoryKey);
 		if (index >= 0) {
 			this._setProp('options', [], index);
 			options.forEach(o => {
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown.js
@@ -65,7 +65,6 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 					// 		key: '',
 					// 		title: '',
 					// 		selected: false,
-					// 		display: true
 					// 	}]
 					// }
 				]
--- a/demo/d2l-filter-dropdown/d2l-filter-dropdown.html
+++ b/demo/d2l-filter-dropdown/d2l-filter-dropdown.html
@@ -61,7 +61,7 @@
 				title: 'Cat ' + i,
 				numSelected: options.filter(o => o.selected).length,
 				options: options
-			}
+			};
 		}
 
 		function search(search, category, data) {
@@ -93,7 +93,7 @@
 					if (option.key === e.detail.optionKey) {
 						option.selected = e.detail.newValue;
 					}
-				})
+				});
 			});
 
 			// Perform search
--- a/demo/d2l-filter-dropdown/d2l-filter-dropdown.html
+++ b/demo/d2l-filter-dropdown/d2l-filter-dropdown.html
@@ -41,7 +41,6 @@
 			<demo-snippet>
 				<template strip-whitespace>
 					<d2l-filter-dropdown class="with-search"></d2l-filter-dropdown>
-					<d2l-filter-dropdown disable-search class="no-search"></d2l-filter-dropdown>
 				</template>
 			</demo-snippet>
 		</div>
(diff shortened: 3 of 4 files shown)


components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -156,6 +156,18 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 		}
 	}
 
+	setFilterOptions(categoryKey, options) {
```
AlexBedley: This function is a convenience function. I could have used `removeFilterOption` to remove every key but this seemed like it deserved it's own helper

components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
```diff
@@ -22,7 +22,7 @@ class D2LFilterDropdownPage extends mixinBehaviors([D2L.PolymerBehaviors.FilterD
 			<d2l-menu label="[[parentTitle]]">
 				<dom-repeat items="[[options]]" as="o">
 					<template>
-					<d2l-menu-item-checkbox text="[[o.title]]" value="[[o.key]]" hidden$="[[!o.display]]" selected=[[o.selected]]></d2l-menu-item-checkbox>
+					<d2l-menu-item-checkbox text="[[o.title]]" value="[[o.key]]" selected=[[o.selected]]></d2l-menu-item-checkbox>
```
AlexBedley: As far as I can tell, `display` was only used by the search stuff so I removed it

components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
```diff
@@ -96,19 +95,19 @@ class D2LFilterDropdownPage extends mixinBehaviors([D2L.PolymerBehaviors.FilterD
...
-		}
-	}
-
-	_caseInsensitiveContainsSubstring(s, sub) {
-		const lowerS = s.toLowerCase();
-		const lowerSub = sub.toLowerCase();
-		return lowerS.indexOf(lowerSub) >= 0;
+		this.dispatchEvent(
+			new CustomEvent(
+				'd2l-filter-dropdown-page-searched',
```
AlexBedley: An interesting thing that could happen now - if we have 2 filter components on the same page, and they both use the same filter category (like might happen to us with the grouped activity view) then searching one will also search the other (but the search text won't be in the other search bar)

AlexBedley:
Ok this is ready for review now. As far as I can tell it works in ie11. No visual changes to any element.

demo/d2l-filter-dropdown/d2l-filter-dropdown.html
```diff
@@ -47,19 +47,66 @@ <h3>Basic d2l-filter-dropdown demo</h3>
...
+				data['Cat' + i] = generateCategory(i);
 			}
-			filter = document.querySelector('d2l-filter-dropdown.no-search');
-			for (i = 0; i < 4; i++) {
-				filter.addFilterCategory('NCat' + i, 'NCat ' + i, i === 0);
-				for (j = 0; j <= i; j++) {
-					filter.addFilterOption('NCat' + i, `Noption${i}-${j}`, `NOption ${i}-${j}`, j === i);
+
+			// Set up filter with our generated data
+			Object.keys(data).forEach(categoryKey => {
```
AlexBedley: The diff in this file is me adding back search functionality to the demo. I am a bit worried that developers might think that this will come with search out-of-the-box. What are your thoughts?

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -79,289 +79,317 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
...
-			assert.isFalse(clearButton.hidden);
-			assert.deepEqual(expected, filter._filters);
-			for (var i = 0; i < filter._filters.length; i++) {
-				var f = filter._filters[i];
-				for (var j = 0; j < f.options.length; j++) {
-					assert.equal(f.options[j].selected, _getOptionByKeys(f.key, f.options[j].key).selected);
-				}
-				assert.equal(expectedNum[f.key], f.numSelected);
-			}
+		suite('unit', () => {
```
AlexBedley: I tried to write integration tests for this function originally but I couldn't figure it out. However, I'm confident that these unit tests prove the function does what it's supposed to.

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -79,289 +79,317 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
...
+					for (var i = 0; i < headers.length; i++) {
+						assert.equal(filter._filters[i].title, headers[i].text);
 					}
-				}
-				done();
+					done();
+				});
 			});
-		});
-		test('changing search input updates option display', function(done) {
```
AlexBedley: Deleted this test because it doesn't make sense anymore

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -79,289 +79,317 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
...
+					done();
+				});
 			});
 		});
 	});
 	suite('d2l-filter-dropdown-page', function() {
 		setup(function() {
 			pageFixture = fixture('page');
 		});
-		test('contains returns true when substring exists', function() {
```
AlexBedley: Since I deleted `_caseInsensitiveContainsSubstring` I also deleted the 4 associated tests

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -79,289 +79,317 @@ import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
...
+
+			pageFixture.addEventListener('d2l-filter-dropdown-page-searched', (e) => {
+				expect(e.composed).to.be.true;
+				expect(e.bubbles).to.be.true;
+				expect(e.detail.value).to.equal(search);
+				expect(e.detail.categoryKey).to.equal(key);
+				done();
+			});
+
+			window.requestAnimationFrame(() => {
```
AlexBedley: I just copied what I saw from this file. Am I doing it right??

demo/d2l-filter-dropdown/d2l-filter-dropdown.html
svanherk: Yeah, I think I'd just log the filter-specific search event to the console, similar to what the `d2l-search-input`'s demo does: https://github.com/BrightspaceUI/inputs/blob/master/demo/d2l-input-search.html#L89.  This is what we do in most UI demos when the consumer is in charge of listening to an event and doing something

svanherk: This would also only work until I did that refactor we discussed offline

components/d2l-filter-dropdown/d2l-filter-dropdown.js
svanherk: Makes sense!  As discussed offline, I want to eventually remove these public methods and just break up the components in a way that they will take the data directly, and if the HM filter's data changes the UI will automatically update, instead of having to be told to update its own copy of the data

test/d2l-filter-dropdown/d2l-filter-dropdown.js
svanherk: This is basically asking for the next UI frame which allows some of the queued operations to proceed.  When people use `window.setTimeout` with a timeout of `0`, they probably wanted to use this.

You may not need this here though, since nothing you've done before this would need time to be applied.

demo/d2l-filter-dropdown/d2l-filter-dropdown.html
svanherk: I'm also seeing weird behaviour if you try to press the Clear button while you have a search applied, then remove the search.  Old applied option state is coming back.  I'd just go with the event!

AlexBedley: > I'm also seeing weird behaviour if you try to press the Clear button while you have a search applied, then remove the search

oh yeah I forgot to hook into the clear button!

> This is what we do in most UI demos when the consumer is in charge of listening to an event and doing something

Bugs aside, I think there is some value in having search work in the demo (for example to see a loading behaviour). At the very least it provides an example of how to implement search. Since I'm already basically done I would lean towards keeping it in. Does that work for you?

svanherk: Yeah that's fine, it's already written and seems like you know the bug fix.  I'm just secretly hoping no one does implement it this way before the refactor (other than the HM one)

components/d2l-filter-dropdown/d2l-filter-dropdown.js
mtjandra: I agree I would say the ideal is it taking data, and would update itself automatically

components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
svanherk: Actually, isn't this needed in order to keep the total counts correct?  To still say you've selected 3 filters, even if you've searched those three away?

mtjandra:
So this would be a breaking change right? Requires major bump? Also IMO I think base and HM wrapper should be in same repo

mtjandra:
Wouldn't this change prevent you from using this component independently as well? It can no longer be used standalone

svanherk:
> Wouldn't this change prevent you from using this component independently as well? It can no longer be used standalone. Would only be useful if someone listen to events and then updates the data correctly

Yeah, but a lot of our UI components are like that.  Button doesn't know what you want it to do when you click it, and the search component doesn't know what type of search you want to do.  HM search, database call, simple javascript filter?

svanherk:
> So this would be a breaking change right? Requires major bump? Also IMO I think base and HM wrapper should be in same repo

This component could be used in a lot of non-HM places though.  I think it eventually belongs in the core UI repo, with the `profile-image-base` component.  And then the HM repos pull in core and extract everything they need.  Whether HM should have a monorepo is also a different, but valid conversation

components/d2l-filter-dropdown/d2l-filter-dropdown.js
svanherk: We could definitely do it as an "in-between", but I'd rather just do the full refactor that avoids passing a huge data object into a component

mtjandra:
> This component could be used in a lot of non-HM places though. I think it eventually belongs in the core UI repo, with the profile-image-base component.

Yeah I was thinking this could be used in a lot of non-HM places as a standalone. Which is why I think it should be able to filter itself without relying on parent. (Also I disagree the `profile-image-base` should be seperated, but that is a different discussion).

> Yeah, but a lot of our UI components are like that. Button doesn't know what you want it to do when you click it, and the search component doesn't know what type of search you want to do. 

Yeah that makes sense, because there is no general meaning to an onClick for a button, and for search you are searching things outside your component ( I am unfamiliar with search so not too sure on this one). 

But here you are searching things within your component. Seems the behaviour is always the same. Maybe there are rare exceptions where you would override the filter from `_caseInsensitiveContainsSubstring` to `_someCustomFilter` but probably very rare.

svanherk:
> But here you are searching things within your component. Seems the behaviour is always the same. Maybe there are rare exceptions where you would override the filter from _caseInsensitiveContainsSubstring to _someCustomFilter but probably very rare.

But that's exactly what we actually need to do for the HM filter, and any use case that needs to support a filter with paging (which I think is the main reason for this change).  Or if they ever want to support searching for data not in the title (for example, searching by course code, but the course title is what is displayed)

AlexBedley:
I agree a little with Miles and a little with Stacey :P I think it's surprising that it can't do search out of the box anymore but I also think that nobody will want client-side searching. I thought about leaving in some default search behavior but I figured it would multiply the testing and YAGNI.

test/d2l-filter-dropdown/d2l-filter-dropdown.js
AlexBedley: - [ ] make sure this is necessary

demo/d2l-filter-dropdown/d2l-filter-dropdown.html
AlexBedley: - [ ] fix clear/search interaction in demo

components/d2l-filter-dropdown/d2l-filter-dropdown-page.js
AlexBedley: - [ ] make sure the total counts are correct

AlexBedley: Ah this exposed a whole range of nasty bugs (we were really relying on having the entire list of options)

svanherk:
A search event is now emitted as part of the changes in https://github.com/BrightspaceUI/facet-filter-sort/pull/30

AlexBedley:
Closing in favour of Stacey's refactor

PR #30, opened 2019-04-23
Refactor to Remove Data Store

svanherk:
This is a huge refactor of the existing component (and will be a major version bump).  Before, the filter UI kept a copy of all the filter category and option states, and also sent events back up so the HM filter could update its copy as well.  This led to a lot of duplicate logic and an inconsistent public API compared to our other UI components.

After this refactor, the filter UI does not keep this data and the component is built up by the client, similar to how dropdowns and menu works.  It sends events back to the client when something has changed, and its up to the client to adjust its data (or make another HM call) as needed.  Updating its data structure will trigger a change to the UI automatically.

Notes:
- The client is now completely in charge of making sure the counts are correct and passed in to be displayed correctly.  Having these components watch their children and update the selected counts was decided would be pretty fragile.
- Because this component has been simplified, there's a lot of deletion in the components and tests.  A lot of that will naturally move to the HM component, or was already there anyways.
- Originally, the plan was also to make a `d2l-filter-dropdown-option` component instead of just using `d2l-menu-item-checkbox`, but it ended up being an exact copy of `d2l-menu-item-checkbox`.  I tried wrapping that component, but that threw off a lot of `menu` code from working, and the wrapper added nothing extra except the component name change.  Thoughts?

Still To Do:
- [x] Add info to the README (currently nothing is there)


svanherk:
I don't think this repo has ever run tests in IE11, I can look into the failures with the sort dropdown component tomorrow

dbatiste:
> Originally, the plan was also to make a d2l-filter-dropdown-option component instead of just using d2l-menu-item-checkbox, but it ended up being an exact copy of d2l-menu-item-checkbox. I tried wrapping that component, but that threw off a lot of menu code from working, and the wrapper added nothing extra except the component name change. Thoughts?

Yeah, the `d2l-menu` code expects direct children to be of type `role="menuitem"` in order to:
* locating the first focusable for when dropdown-menu is opened
* manage focus (alphanumeric keys, arrow keys)
* styling of the menu and items
* consideration of hidden menu-items for focus and styling

If the `role="menuitem"` is buried deeper, or in a shadow-DOM, these things become quite messy.  To enable different visual layouts, and also different menu-items roles (normal, checkbox, radio), the menu-item code was factored into a behavior.  While it would be possible to create an implementation (using that behavior) merely for the purpose of having a certain tag-name, that wasn't the original intention.  I don't have a super strong position on whether we should/shouldn't do that, but the menu-item should be a direct child.

svanherk:
> While it would be possible to create an implementation (using that behavior) merely for the purpose of having a certain tag-name, that wasn't the original intention. I don't have a super strong position on whether we should/shouldn't do that, but the menu-item should be a direct child.

Yeah exactly, because of the behaviour I could easily make a new type of `menu-item`, but for now it'd be basically a copy.  Maybe this would be good to do, if we thought we'd change the design for filter components in the future, or if we don't want all changes to `d2l-menu-item-checkbox` to automatically be picked up by everyone using a filter component.  If we don't expect those to be a problem, seem a bit like overkill for _just_ the name change. Hmm... 🤔

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -0,0 +1,163 @@
...
+ */
+
+class D2LFilterDropdownCategory extends mixinBehaviors([D2L.PolymerBehaviors.Tabs.TabPanelBehavior, D2L.PolymerBehaviors.FilterDropdown.LocalizeBehavior], PolymerElement) {
+	static get template() {
+		return html`
+			<style include="d2l-tab-panel-styles">
+				:host {
+					margin: -1px -1rem 0 -1rem;
+					border-top: 1px solid var(--d2l-color-gypsum);
+					padding: 1.2rem 1rem 0 1rem;
```
dlockhart: Looks like some margins and padding that will need to be reversed in RTL.

demo/d2l-filter-dropdown/d2l-demo-filter.js
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * @customElement
+ * @polymer
+ */
+
+class D2LDemoFilter extends PolymerElement {
+	static get template() {
+		return html`
+			<style>
```
dlockhart: Not sure if this should be `display: block` or `inline-block`? You should only need to specify the `[hidden]` CSS if the `display` varies from the default of `inline`.

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -1,387 +1,155 @@
-import '@polymer/iron-test-helpers/mock-interactions.js';
 import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
 
 (function() {
 	var filter;
```
AlexBedley: nit: this should be scoped to the individual `suite`. Not a problem currently but could cause a nasty bug. Same with `categories`.

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -1,387 +1,155 @@
...
-					assert.equal(pages[i].options.length, opts.length);
-					for (var j = 0; j < opts.length; j++) {
-						assert.equal(pages[i].options[j].title, opts[j].text);
-					}
-				}
-				done();
-			});
+	});
+
+	suite('d2l-filter-dropdown-category', function() {
```
AlexBedley: nit: this deserves its own file for searchability (looking at the file system you'd assume dropdown-category has no tests)

test/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -1,387 +1,155 @@
...
-			_getExpectedAndImport(filter);
-			var removal = categories[0].key;
-			filter.removeFilterCategory(removal);
-			window.requestAnimationFrame(function() {
-				var headers = _getTabPanels();
-				assert.equal(categories.length - 1, headers.length);
-				assert.deepEqual(categories.filter(c => c.key !== removal).map(c => c.title), [].map.call(headers, h => h.text));
-				done();
-			});
+		test('setting the search-value aattrbiute updates the value in the search input', function() {
```
AlexBedley: nit: if you do change this file, might as well correct the spelling mistake :P but otherwise don't bother

AlexBedley:
> * Originally, the plan was also to make a `d2l-filter-dropdown-option` component instead of just using `d2l-menu-item-checkbox`, but it ended up being an exact copy of `d2l-menu-item-checkbox`.  I tried wrapping that component, but that threw off a lot of `menu` code from working, and the wrapper added nothing extra except the component name change.  Thoughts?

I'm fine with the current state, especially if the wrapper is causing more problems than it solves.

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
svanherk: The left and right are the same for both, it's just the top and bottom that needed to be customized

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -0,0 +1,163 @@
...
+				d2l-menu {
+					margin: 0 -1rem;
+					width: auto;
+				}
+			</style>
+			<div class="d2l-filter-dropdown-page-search" hidden$="[[disableSearch]]">
+				<d2l-input-search placeholder="[[localize('searchBy', 'category', categoryText)]]" value="[[searchValue]]"></d2l-input-search>
+			</div>
+			<d2l-menu label="[[text]]">
+				<slot class="d2l-menu-items"></slot>
```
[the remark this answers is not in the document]
svanherk: It isn't named, it just has a class on it that isn't actually needed.  I was experimenting with something that didn't work and forgot to remove, will delete it.

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -0,0 +1,163 @@
...
+	static get observers() {
+		return [
+			'_updateTabText(categoryText, selectedOptionCount)'
+		];
+	}
+
+	ready() {
+		super.ready();
+		this._handleSearchChange = this._handleSearchChange.bind(this);
+		this._handleMenuItemChange = this._handleMenuItemChange.bind(this);
```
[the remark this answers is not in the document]
svanherk: It's needed because I'm attaching the listener to the `menu` element instead of to `this`, but I want access to `this.key` in `_handleMenuItemChange`

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -0,0 +1,163 @@
...
+						selected: e.detail.selected
+					},
+					composed: true,
+					bubbles: true
+				}
+			)
+		);
+	}
+
+	_dispatchSelected() {
```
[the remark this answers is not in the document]
svanherk: It's overriding the function of the same name in https://github.com/BrightspaceUI/tabs/blob/master/d2l-tab-panel-behavior.js#L74

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -0,0 +1,163 @@
...
+					bubbles: true,
+					composed: true
+				}
+			)
+		);
+	}
+
+	_updateTabText(text, numSelected) {
+		if (numSelected === 0) {
+			this.text = text;
```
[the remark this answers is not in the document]
svanherk: It already is in https://github.com/BrightspaceUI/tabs/blob/master/d2l-tab-panel-behavior.js#L59

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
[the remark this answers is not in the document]
AlexBedley: Ah well that's confusing... as well as the "list text in properties" comment below. I guess there's no js way to handle it like explicit implementation in c#. Is it worth a comment?

svanherk:
Sorry for the delay, been heads don in `my-courses` defects.  Addressed code review comments and added some README comments and images

svanherk:
Any more thoughts here?

README.md
```diff
@@ -5,49 +5,106 @@
...
-        font-size: 20px;
-        font-family: 'Lato', 'Lucida Sans Unicode', 'Lucida Grande', sans-serif;
-      }
-    </style>
-    <next-code-block></next-code-block>
-  </template>
-</custom-element-demo>
-```
--->
+Then, add the `d2l-filter-dropdown` as the top level filter component.  For each filter category, add a `d2l-filter-dropdown-category` component, which is a custom `d2l-tab-panel` that includes a `d2l-menu`.  Then, for each filter option in that category, you should use the `d2l-menu-item-checbox` component.  For example:
```
AdmiralCoco: Typo: `d2l-menu-item-chekbox` -> `d2l-menu-item-checkbox`

svanherk:
@AlexBedley This will be a major bump and I don't think anyone else in BSI is using this repo, so the update to `d2l-hm-filter` doesn't have to be immediate and if there are any issues we can come back here and adjust.

PR #31, opened 2019-04-25
Fix ie11 edge issues

AdmiralCoco:
- Fix tests (issues with event bubbling)
- Fix CSS selectors (rtl applying when not rtl in Edge)
- Minor cleanup


components/d2l-sort-by-dropdown/d2l-sort-by-dropdown.js
```diff
@@ -107,15 +106,16 @@ class SortByDropdown extends mixinBehaviors(
 		super.connectedCallback();
 		this._boundListeners = { _onItemSelect: this._onItemSelect.bind(this) };
 		afterNextRender(this, function() {
-			this.$['d2l-sort-by-dropdown-menu'].addEventListener('d2l-menu-item-change',
+			// Issues with event bubbling in Edge/IE11 - need to listen directly on the menu
+			this.$['d2l-sort-by-menu'].addEventListener('d2l-menu-item-change',
```
AdmiralCoco: The event fired by `d2l-sort-by-dropdown-option` was not being caught by the event listener, but only during tests. Didn't see any issues on the demo page

 I am guessing it's related to event bubbling 🤷‍♂️

PR #32, opened 2019-05-09
Add properties to set header and opener langterms

emil-furniss:
We're using `d2l-filter-dropdown` in Activity Feed and are loving it. However, we are also using it as a `Send To` selector, for which we to modify the terminology a little. We also don't have a need to display how many selections have been made in the opener button, as we're displaying tags alongside it which show exactly which selections have been made.

This PR introduces new properties that allow the consumer to provide their own alternative terms for
- `filterBy` by setting the property `headerText`
- `filter` by setting the property `openerText`
- `filterSingle` by setting the property `openerSingle`
- `filterMultiple` by setting the property `openerMultiple`

as well as the option to disable displaying different terms for different number of selections on the opener button via `disable-opener-text-variation`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,17 @@ Then, add the `d2l-filter-dropdown` as the top level filter component.  For each
 </d2l-filter-dropdown>
 ```
 
+The default lang terms can be overidden by setting the appropriate attributes.
+
+```html
+<d2l-filter-dropdown total-selected-option-count="3" header-text="Send To" opener-text="Send" opener-single="Sending To One" opener-multiple="Sending To Many">
+	<d2l-filter-dropdown-category key="1" category-text="Category 1" selected-option-count="1">
+		<d2l-menu-item-checkbox selected text="Option 1 - 1" value="1"></d2l-menu-item-checkbox>
+		<d2l-menu-item-checkbox text="Option 1 - 2" value="2"></d2l-menu-item-checkbox>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -52,13 +52,31 @@ Then, add the `d2l-filter-dropdown` as the top level filter component.  For each
...
 #### Attributes
 
 `d2l-filter-dropdown`:
 
 - `max-width` (optional): Sets the max-width of the filter dropdown.
 - `min-width` (optional): Sets the min-width of the filter dropdown.
 - `total-selected-option-count`: The total number of selected filter options across all categories.  When options are selected and de-selected, the consumer is responsible for updating this number after updating its own data store.
+- `header-text` (optional): Sets the text for the filter content header.
+- `opener-text` (optional): Sets the text for the opener when there are no selections.
+- `opener-single` (optional): Sets the text for the opener when there is a single selection.
```
svanherk: Would be wordier, but I think this would be clearer as `opener-text-single` and `opener-text-multiple`.

svanherk:
All of these additions seem like a good idea to me!  I'm a little torn about the case where they provide their own text, then they lose out on the number being included automatically.  But they may not want that or they may want it written out, and it's not a lot of work for them to update it when a selection is made since they need to update the count at the same time. So I think this is the way to go, and the general case of filter is still working as before.

PR #38, opened 2019-10-07
Add d2l-filter-dropdown-option to master

svanherk:
Bringing https://github.com/BrightspaceUI/facet-filter-sort/pull/36 into the latest master version, and adjusting the naming, demo, tests and readme accordingly.

This will need to be a major version jump due to the event name change.

### DIFF
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown-option.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown-option.js
@@ -1,0 +1,49 @@
+import '@polymer/polymer/polymer-legacy.js';
+
+import 'd2l-icons/d2l-icons.js';
+import 'd2l-menu/d2l-menu-item-selectable-styles.js';
+import 'd2l-menu/d2l-menu-item-selectable-behavior.js';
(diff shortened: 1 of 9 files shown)


components/d2l-filter-dropdown/d2l-filter-dropdown-option.js
```diff
@@ -0,0 +1,49 @@
...
+const $_documentContainer = document.createElement('template');
+
+$_documentContainer.innerHTML = `<dom-module id="d2l-filter-dropdown-option">
+	<template strip-whitespace="">
+		<style include="d2l-menu-item-selectable-styles">
+			:host > span {
+				white-space: normal;
+			}
+		</style>
+		<d2l-icon icon="d2l-tier1:check" aria-hidden="true"></d2l-icon>
```
dbatiste: I am surprised the `aria-hidden` is needed here. (?)

svanherk: Discussed offline - going to leave for now in case there's an accessibility reason, but we should revisit whether this is really needed when we convert to Lit.  FYI @margaree

PR #42, opened 2020-02-29
allow no update of option checked state

mdulat:
To allow `d2l-filter-dropdown-option` to not update the checked state automatically, instead allowing consumer to set `selected` explicitly. This allows consumer components to not uncheck a menu item when it is clicked multiple times (use case is when one item must always be ckecked).

### DIFF
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown-option.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown-option.js
@@ -31,6 +31,10 @@ Polymer({
 		'role': 'menuitemcheckbox'
 	},
 
+	properties: {
+		noUpdate: Boolean
+	},
+
 	attached: function() {
 		afterNextRender(this, function() {
 			this.listen(this, 'd2l-menu-item-select', '_onSelect');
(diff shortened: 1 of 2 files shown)


dbatiste:
This seems like it could be confusing to screen reader users.  It is kinda like it's behaving as disabled.  If the menu item behavior support disabled however, that could also be confusing to them, because for the use case mentioned, the disabled state would be changing as they selected/unselect items.  🤔

mdulat:
> This seems like it could be confusing to screen reader users. It is kinda like it's behaving as disabled. If the menu item behavior support disabled however, that could also be confusing to them, because for the use case mentioned, the disabled state would be changing as they select/unselect items. 🤔

Ah, you're right, I think we want radio items in our case here instead of checkboxes. We're replacing the dropdowns in the question library import page (see https://qa2020420950g.bspc.com/d2l/lms/qc/main_frame.d2l?ctid=1&ti=51000&ou=123117, select Import --> Browse Existing Questions). Sort is replaced with `d2l-sort-by-dropdown`, Filter with `d2l-filter-dropdown`, and Source is kind of a combination of the two, because it should be a menu with radio buttons, but it also needs tabs, so we used `d2l-filter-dropdown`. Hmmm.

mdulat:
Turns out we can use `d2l-sort-by-dropdown-option` within `d2l-filter-dropdown`, does exactly what we want.

PR #43, opened 2020-03-19
Added a slot to filter-dropdown

EvanLemieux:
Used for https://github.com/Brightspace/folio-app/pull/451


components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -48,6 +48,7 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 					<d2l-tabs>
 						<slot></slot>
 					</d2l-tabs>
+					<slot name="footer" slot="footer"></slot>
```
jstefaniuk-d2l: Context: this allows consumers of the filter dropdown to access the footer slot of the inner dropdown component if they need

PR #48, opened 2020-04-24
Switch things to core

dbatiste:
This PR probably looks scarier than it should.  It's basically house-keeping and couple conversions to Lit in order to remove dependency on Polymer behaviors.

* replace `d2l-filter-dropdown-category` custom Polymer tab panel with Lit version
* replace `d2l-filter-dropdown-option` custom Polymer menu item with Lit version
* replace `d2l-sort-by-dropdown-option` custom Polymer menu item with Lit version
* update all imports to pull from `core` except for typography where mixin is being used
* cleanup all demo pages so they are easier to work with and don't use iron demo snippet
* fix a few tests that broke as a result of Lit rendering being async


dbatiste:
Cool, yeah, I tried to keep the commits targeted the specific changes for anyone looking.  Glad it helped.  :)

PR #49, opened 2020-05-02
Applied filters

jstefaniuk-d2l:
A multi-select-list allowing the user to see (and remove) the currently applied filters

(example showing use in portfolio) 
image


components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -0,0 +1,218 @@
...
+		this._setFilter();
+		const dropdown = this._target;
+
+		if (!dropdown) {
+			return [];
+		}
+
+		const results = {};
+		[...dropdown.children]
+			.filter(x => x.nodeName === DROPDOWN_CATEGORY_NAME)
```
ctwomey1, line 174: Is filtering all children going to be an issue.  In portfolio at least we're probably going to have an arbitrary number of tags to apply right?

Could this be a case where state management is warranted?

jstefaniuk-d2l: this is only ran when slots are changed on the filter menu, not the multi-select, so in portfolio it runs once with elements there (plus a couple more times when there's no children)

ctwomey1: Right but if you have a scenario of:

- applied filters and filter menu load (empty)
- Async process to load filters happens
- Each filter is slotted sequentially
- applied filters slotted fires N times iterating over the full children collection of filters

Or is there something smart enough to dedupe those interactions?

jstefaniuk-d2l: It's not too hard to only check the changed category, so we could restrict it to only rechecking the category that changed

svanherk:
Sorry for the delay, I can take a look this afternoon!  Usually we wouldn't want a `labs` component used in a BrightspaceUI component, but `facet-filter-sort` is really a labs repo at heart and we just haven't had time to move it over.  Is there a design for this? There's a "to do" item to get filtering an actual design on design.d2l, and would be good to consider this at the same time.

svanherk:
You're planning to use this in an FRA, correct?  BSI is on an older version of this repo and work is needed to get it up-to-date.  I want to use this in my-courses so that work will be happening _soon_, but not immediately...

jstefaniuk-d2l:
We're in a FRA so no need to worry about BSI!

Fisayo has been working on the design for this:
[image](https://user-images.githubusercontent.com/14077760/81327759-62151e00-9061-11ea-8e6b-3965d26afa5d.png)

The "filter tag" styling is still a bit in-flux, but the general gyst is a multi-select that reflects the contents of the filter dialog

If I recall correctly another designer was interested in using the same component for their project as well, for quickEval _I think_

svanherk:
> We're in a FRA so no need to worry about BSI!

Ok that's good, no delay then

> If I recall correctly another designer was interested in using the same component for their project as well, for quickEval I think

If it is Quick Eval, that's the place using the old version (https://github.com/BrightspaceHypermediaComponents/common/blob/master/components/d2l-hm-filter/d2l-hm-filter.js needs to be updated).  So they'd need to upgrade to use it, which is probably a nice inventive if I don't get to updating it first.

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -153,7 +157,7 @@ class FilterDropdownCategory extends LocalizeStaticMixin(TabPanelMixin(LitElemen
 				</d2l-input-search>
 			</div>
 			<d2l-menu @d2l-menu-item-change="${this._handleMenuItemChange}" label="${this.text}">
-				<slot></slot>
+				<slot @slotchange="${this._onSlotChange}"></slot>
```
svanherk, line 160: I'm assuming you don't need to support IE11 or that this isn't going out in the June release?

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -144,6 +144,10 @@ class FilterDropdownCategory extends LocalizeStaticMixin(TabPanelMixin(LitElemen
 		});
 	}
 
+	_onSlotChange() {
+		this.dispatchEvent(new CustomEvent('d2l-filter-dropdown-category-slotchange', { bubbles: true, composed: true, detail: { categoryKey: this.key } }));
```
svanherk, line 148: I think you should be able to set this to `composed: false`, since you add the listener directly onto the `d2l-filter-dropdown` and `d2l-filter-dropdown-category` should always be in the same shadowRoot.  Nothing above that should need to know about a slot change.  A lot of the other events here also probably don't need it, but we're trying to be more aware of it going forward

components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
jstefaniuk-d2l: I don't think we've been supporting ie11 in portfolio, since we're using slotchange elsewhere, maybe I can leave a comment in the readme `applied-filters` section saying that it uses slotchange and won't support ie

svanherk: Yeah a comment would be good, may be unlikely it will start being used elsewhere in the next week and a half but you never know

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -0,0 +1,227 @@
...
+			_entries: { type: Array },
+			_selectedEntries: { type: Array },
+			_target: { type: Object },
+		};
+	}
+
+	static get styles() {
+		return [bodyCompactStyles, css`
+			:host {
+				display: block;
```
svanherk, line 26: Minor: For web components with a default `display` that's different than `inline`, it's [good practice](https://developers.google.com/web/fundamentals/web-components/best-practices#add-a-:host-display-style-that-respects-the-hidden-attribute.) to add handling for the `hidden` attribute:
```
:host([hidden]) {
  display: none
}
```

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -0,0 +1,227 @@
...
+			d2l-labs-multi-select-list {
+				flex: 1;
+			}
+
+			.d2l-applied-filters-wrapper {
+				display: flex;
+			}
+
+			.d2l-applied-filters-applied-filters-label {
+				margin-right: 0.25rem;
```
svanherk, line 38: We probably need to handle the RTL case here and include https://github.com/BrightspaceUI/core/blob/master/mixins/rtl-mixin.js

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	_filterOptionChanged() {
+		this._setSelectedOptions();
+	}
+
+	_multiSelectItemDeleted(e) {
+		const entryIndex = Number.parseInt(e.path[0].attributes.index.value);
+		const entry = this._selectedEntries[entryIndex];
+		entry.dispatchEvent(new CustomEvent('d2l-menu-item-select'));
```
svanherk, line 129: What would the consumer use this event for?

jstefaniuk-d2l: The deletion was for the multi-select list, I put up another commit to not do this with an event since it was unintuitive

svanherk:
For the Edge tests, you may be running into this issue: https://github.com/Polymer/tools/pull/727
They've merged the fix, but have yet to release it (https://github.com/Polymer/tools/blob/master/packages/build/CHANGELOG.md#change-log).  As you can see in the issue thread, we've asked multiple times, but there may never be another release.  We're trying to move everything off of `web-component-tester`, but that's not a super quick change.  But maybe you can pull their change in to see if that's your issue?

jstefaniuk-d2l:
Hmm that seems like it may be the issue, although I don't think we pull in polymer tools here since it's using the wct-browser-legacy module

The tests all work if you refresh them enough times, and were flaky in edge before this pr, so I may just disable edge tests for now until this gets the new test framework (unless you have any other ideas, or think I should take further action)

svanherk:
We've had to disable Edge in a number of other repos because of this.  It's not great, and it has led to issues (something came up in #web-platform the other day that would have been caught if legacy Edge tests were running in `activities`).  Ideally we start using the recommendations of the `open-wc` group (https://open-wc.org/testing/) as soon as possible so we can turn Edge testing back on, and do manual old Edge checks in the meantime

jstefaniuk-d2l:
@svanherk thanks for the help!

PR #53, opened 2020-06-24
Add custom-elements.json

margaree:
This PR adds web components analyzer to generate custom-elements.json and adds more jsdoc comments throughout the components.
I did add custom-elements.json into source control since this isn't published to npm so I couldn't do what was done in core. It would be great to have this more automated so if anyone has suggestions there, that would be awesome.

### DIFF
--- a/components/d2l-applied-filters/d2l-applied-filters.js
+++ b/components/d2l-applied-filters/d2l-applied-filters.js
@@ -11,9 +11,15 @@ const DROPDOWN_NAME = 'D2L-FILTER-DROPDOWN';
 const DROPDOWN_CATEGORY_NAME = 'D2L-FILTER-DROPDOWN-CATEGORY';
 const DROPDOWN_OPTION_NAME = 'D2L-FILTER-DROPDOWN-OPTION';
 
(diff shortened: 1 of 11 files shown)


svanherk:
Would it be crazy to add this to the frau-ci auto-bumping? During the `[skip ci]` step that ups the version, could it look for a certain `npm` command and/or the existence of the file, and run it and commit the new file?

svanherk:
Or if we're moving away from `frau-ci` for UI and Labs, can we start moving everyone to the semver stuff and work it into that workflow? Or does that only make sense for components publishing to npm?

margaree:
So I think many labs components (at least all the ones I looked at quickly) are publishing to npm so I think for this custom-elements.json, adding the workflow used in core into them would make the most sense rather than adding it into `frau-ci`.

dlockhart:
Let's not add `frau-ci` to anything new -- `semantic-release` would work well here as it can do GitHub and/or NPM releases. In the configuration for updating `package.json`, you can include other files that should be committed, so we could just add `custom-elements.json`.

PR #55, opened 2020-07-31
DE39995 - Fix d2l-filter-dropdown-category-selected event

svanherk:
The `d2l-filter-dropdown-category-selected` event has not been firing since `_dispatchSelected` was accidentally removed from the `d2l-tab-panel-mixin`.  This feels better than relying on a private function from the mixin anyways.

Tests would have caught this, but because we use npm caches it did not.  I've turned those caches off.

### DIFF
--- a/components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
+++ b/components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
@@ -165,6 +165,9 @@ class FilterDropdownCategory extends LocalizeStaticMixin(TabPanelMixin(LitElemen
 			this.selected = !this.selected;
 			this.__onSelect(e);
 		});
+		this.addEventListener('d2l-tab-panel-selected', () => {
+			this._dispatchSelected();
+		});
 	}
 
 	_onSlotChange() {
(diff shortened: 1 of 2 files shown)


components/d2l-filter-dropdown/d2l-filter-dropdown-category.js
```diff
@@ -165,6 +165,9 @@ class FilterDropdownCategory extends LocalizeStaticMixin(TabPanelMixin(LitElemen
 			this.selected = !this.selected;
 			this.__onSelect(e);
 		});
+		this.addEventListener('d2l-tab-panel-selected', () => {
```
svanherk, line 168: I like the idea of stopping propagation of this event, but that doesn't happen currently and someone could be relying on that, so leaving for now.

dbatiste: The [d2l-dropdown-tabs](https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-tabs.js#L29) component appears to depend on this event.

svanherk: Which makes sense, we don't want it to rely on a specific filter event.  It's the only use-case for now, but there may be more in the future

PR #57, opened 2020-08-06
"Clear filters" button added, below the list of filters in Applied Filters component

adamsk34:
https://rally1.rallydev.com/#/detail/userstory/406501674732?fdp=true

Added functionality:
* When 4 or more filters are applied, a "Clear filters" button appears below the list of applied filters
* When 3 or fewer filters are applied, the "Clear filters" button disappears
* Clicking "Clear filters" removes all applied filters
* Accessibility:
    * When entering the "Clear filters" button, NVDA says "clear filters button"
    * When clicking on "Clear filters", NVDA says "all filters removed" and focus moves from the "Clear filters" button (which just disappeared) and is put on the filter dropdown button

## Update

The clear filters button has recently moved to be below the list of applied filters since the last time this PR was reviewed.

lrt
cf_1

rtl
cf_2

## Testing

For testing, there is a problem with the demo where the dropdown does not get updated when you click "Clear filters". This is a problem with the demo's dropdown and not with "Clear filters". The "Clear" button in the dropdown has the same problem in the demo.

To be sure that "clear filters" is working, this change should also be tested in folio-app, because there the dropdown is working correctly. Test on ltr and rtl.


components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -107,9 +110,13 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 
 	constructor() {
 		super();
+		this._clearFiltersClicked = this._clearFiltersClicked.bind(this);
 		this._clearSelected = this._clearSelected.bind(this);
+		this._handleDropdownChange = this._handleDropdownChange.bind(this);
 		this._setSelectedOptions = this._setSelectedOptions.bind(this);
 		this._update = this._update.bind(this);
+
+		this._hideClearFiltersButton = true;
```
adamsk34, line 119: This line means that the “clear filters” button does not appear when the page first loads. I have assumed that the page will never load with 4+ filters already applied.

jstefaniuk-d2l: If we handle the hiding at render time then we won't have to worry about this edge case

```
<d2l-button-subtle text="${this.localize('clearFilters')}" ?hidden="${this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD}" @click="${this._clearFiltersClicked}"></d2l-button-subtle>
```
but you could also go one step further and do something like
```
shouldHideClearButton() {
}
...
${this.shouldHideClearButton(this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD) ? html`<d2l-button-subtle text="${this.localize('clearFilters')}" @click="${this._clearFiltersClicked}"></d2l-button-subtle>` : ''}
```

Doing this logic at render is good because it can be evaluated purely from other properties

adamsk34: I see what you're saying. However, `this._selectedEntries` is undefined at render time, so there would have to be another solution

adamsk34: I think I have a good solution, I'll push in a minute

adamsk34: I tested and pushed my change. Now "clear filters" will appear on page render if there are enough filters already applied.
I renamed my function `this._handleDropdownChange` to `this._refreshFilters` and I call it in the constructor

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -209,12 +216,29 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
...
 	}
 
 	_setSelectedOptions() {
 		this._selectedEntries = [].concat.apply([], Object.values(this._entries || {})).filter(x => x.selected);
 	}
 
+	_refreshFilters() {
+		this._setSelectedOptions();
+		this._hideClearFiltersButton = this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD;
```
jstefaniuk-d2l, line 228: this property probably not needed any more

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -209,12 +216,29 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 
 	_clearSelected() {
 		this._selectedEntries = [];
+		this._hideClearFiltersButton = true;
```
marcsoiferman, line 219: For consistency, I would set this to `this._hideClearFiltersButton = this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD;`

If a change to design is ever made where we always want to show the button, the obvious change would be to change CLEAR_FILTERS_THRESHOLD to something like -1, but this would still cause the button to disappear

components/d2l-applied-filters/d2l-applied-filters.js
adamsk34: Good point

components/d2l-applied-filters/d2l-applied-filters.js
marcsoiferman: Actually it looks like you just put the condition straight into the control, so you can probably just remove this line

adamsk34: Good point. However I'm actually going to remove `_hideClearFiltersButton` all together, Jeff pointed out that it's not really needed anymore

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -237,6 +261,8 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 			<div class="d2l-applied-filters-wrapper">
 				<span id="d2l-applied-filters-label" class="d2l-applied-filters-applied-filters-label d2l-body-compact">${this.localize('appliedFilters')}</span>
 				${filters}
+				<!-- <d2l-button-subtle text="${this.localize('clearFilters')}" ?hidden="${this._hideClearFiltersButton}" @click="${this._clearFiltersClicked}"></d2l-button-subtle> -->
```
marcsoiferman, line 264: Just a reminder so that you don't forget to remove commented out code before finalizing this

adamsk34: Good catch, I added that just now by accident. I removed it

components/d2l-applied-filters/d2l-applied-filters.js
adamsk34: I got rid of `_hideClearFiltersButton `, in the process I also removed the `_refreshFilters` function

components/d2l-applied-filters/d2l-applied-filters.js
adamsk34: I got rid of `_hideClearFiltersButton`, in the process I also removed the `_refreshFilters` function

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -237,6 +254,7 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 			<div class="d2l-applied-filters-wrapper">
 				<span id="d2l-applied-filters-label" class="d2l-applied-filters-applied-filters-label d2l-body-compact">${this.localize('appliedFilters')}</span>
 				${filters}
+				<d2l-button-subtle text="${this.localize('clearFilters')}" ?hidden="${this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD}" @click="${this._clearFiltersClicked}"></d2l-button-subtle>
```
jstefaniuk-d2l, line 257: It's generally worth adding this rule to the css when you use `hidden`, without it there are ways for the component to be "hidden" but still show:
```
			[hidden] {
				display: none !important;
			}
```

Also PSA, this is probably one of the only places you can use the `!important` keyword without me asking you to remove it in code review 😛

components/d2l-filter-dropdown/d2l-filter-dropdown.js
```diff
@@ -141,7 +141,7 @@ class D2LFilterDropdown extends mixinBehaviors([D2L.PolymerBehaviors.FilterDropd
 		this.removeEventListener('d2l-dropdown-close', this._handleDropdownClose);
 	}
 
-	_clearFilters() {
+	clearFilters() {
```
marcsoiferman, line 144: I think we should leave the underscore here to indicate that this is supposed to be an internal method.  @jstefaniuk-d2l can pitch in

jstefaniuk-d2l: We switched to using this externally here so that the applied filters component can clear the dropdown filters: https://github.com/BrightspaceUI/facet-filter-sort/pull/57/files#diff-53571d5ed7be88c8a70d36ce8872776dR227

components/d2l-applied-filters/d2l-applied-filters.js
adamsk34: I added the css, I assumed the right place to put it was in the `styles` function

components/d2l-filter-dropdown/d2l-filter-dropdown.js
jstefaniuk-d2l: The alternative would be having the client handle "applied-filters-clear-filters" and "dropdown-clear-filters" separately, which may not be the worst thing in the world, but this seemed slightly more convenient

jstefaniuk-d2l:
I think once the code is in a stable place, a few tests could be added for this feature: https://github.com/BrightspaceUI/facet-filter-sort/blob/master/test/d2l-applied-filters/d2l-applied-filters.html

adamsk34:
In my recent changes I wrote tests for clear filters and added the id `d2l-clear-filters-button` to the clear filters button so I could test it easily.
Note: in the file `test/d2l-applied-filters/d2l-applied-filters.html` there is html at the top of the page that has 3 filters selected when the page first renders. That's relevant when reviewing the tests I wrote

test/d2l-applied-filters/d2l-applied-filters.html
```diff
@@ -101,6 +101,107 @@
 						done();
 					});
 				});
+
+				test('clear filters button is hidden when 3 filters are applied', function() {
```
adamsk34, line 105: There is html at the top of this file that makes it so 3 filters are selected when the page first renders

test/d2l-applied-filters/d2l-applied-filters.html
```diff
@@ -101,6 +101,107 @@
...
+
+					options[0].deselect();
+					options[3].deselect();
+
+					afterNextRender(options[0], () => {
+						assert.equal(clearFilters.hidden, true);
+						done();
+					});
+				});
+				test('selecting a 4th filter & clicking clear filters fires a d2l-filter-dropdown-option event', function(done) {
```
jstefaniuk-d2l, line 174: wrong event name

test/d2l-applied-filters/d2l-applied-filters.html
```diff
@@ -101,6 +101,107 @@
...
+						done();
+					});
+
+					options[1].click();
+
+					afterNextRender(clearFilters, () => {
+						clearFilters.click();
+					});
+				});
+				test('selecting a 4th filter & clicking clear filters removes filters & hides clear filters button', function(done) {
```
jstefaniuk-d2l, line 188: Question, should we actually remove filters in this component, regardless of the filters in the data source? Or should we let the consumer 100% handle it?

test/d2l-applied-filters/d2l-applied-filters.html
adamsk34: Fixed

test/d2l-applied-filters/d2l-applied-filters.html
adamsk34: I hadn't considered letting the consumer handle it themselves. What is the data source for the filters? For us would that be the variables in folio-app that hold the list of applied filters?

jstefaniuk-d2l: The data source for the filters is essentially the dropdown. When you send the event in folio, I bet that it'll clean up this component without any code in here doing so (worth a try!)

The reason for doing it that way is otherwise the dropdown and the applied filters component could potentially get out-of-sync, which could be confusing for the user

adamsk34: There is code (that I didn't write) in d2l-applied-filters.js that clears the filters in the applied filters component when the `d2l-filter-dropdown-cleared` event is fired
```
target.addEventListener('d2l-filter-dropdown-cleared', this._clearSelected, true);
```
What my code was doing (until I changed it a few minutes ago) was calling `this._clearSelected` to clear the applied filters component when clear filters is pressed, but it was doing that right before firing the `d2l-filter-dropdown-cleared` so I removed my redundant function call

adamsk34: If we change it so the consumer handles the clearing of the applied filters component, we would have to use a different event than `d2l-filter-dropdown-cleared`

jstefaniuk-d2l: Interestinggg, what's here is likely good then :+1:

jstefaniuk-d2l:
Code LGTM: I think this is ready for someone to test in folio

adamsk34:
> 
> 
> Code LGTM 👍 I think this is ready for someone to test in folio

Nice 🎉

Something I should note here for anyone to read is that, in the demo for facet-filter-sort, the filter dropdown behaves weird because it isn't listening for the `d2l-filter-dropdown-cleared` event like folio-app is. It's worth it to try out the facet-filter-sort demo anyway but whoever is testing the "clear filters" button should for sure test it out in folio-app

jameswklassen:
Not sure if this is due to ambiguity in the spec, or something I missed.

Is the clear button supposed to align to the right, or is it supposed to appear as an element appended to the list?

Screen Shot 2020-08-12 at 8 11 42 AM

adamsk34:
@jameswklassen that's due to ambiguity in the spec. The button should be right aligned, I have confirmed this with Fisayo. I'll update the issue to make that clear

adamsk34:
I am wrong, the "clear filters" button should actually appear as an element in the list. Thank you for catching that! I will make the change

jameswklassen:
A minor thing, it looks like the vertical alignment of the elements is a little wacky. Although this might be fixed with your other visual changes, so probably not worth looking at until then.

Screen Shot 2020-08-12 at 4 48 48 PM

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -60,9 +61,22 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
...
+			#d2l-list-holder {
+				flex: 1;
+			}
+
 			:host([hidden]) {
 				display: none;
 			}
+
+			[hidden] {
+				display: none !important;
```
marcsoiferman, line 78: What does !important mean?

adamsk34: It means that other CSS can't override `display: none`. Adding this line was recommended to me by Jeff as one of the rare cases where `!important` is not unacceptable.

This doc

adamsk34: It means that other CSS can't override `display: none`. Adding this line was recommended to me by Jeff as one of the rare cases where `!important` is not unacceptable. It's acceptable in this case because we are hiding a component and we really don't want another part of the page deciding that it should be visible. `!important` is generally a bad practice because it voilates the natural cascading in your stylesheets. I think of it kind of like `GOTO` but for CSS.

@jstefaniuk-d2l can let me know if I have accurately explained the reasoning.

This doc explains it under the "The !important exception"
https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

apalaniuk: It disregards concepts like specificity to ensure this declaration is always the one used when an element has the `hidden` attribute.

test/d2l-applied-filters/d2l-applied-filters.html
```diff
@@ -101,6 +101,107 @@
...
+						done();
+					});
+				});
+				test('selecting a 4th filter & clicking clear filters fires a d2l-filter-dropdown-cleared event', function(done) {
+					const clearFilters = appliedFilters.shadowRoot.querySelector('#d2l-clear-filters-button');
+					const options = dropdown.querySelectorAll('d2l-filter-dropdown-option');
+
+					document.addEventListener('d2l-filter-dropdown-cleared', () => {
+						done();
+					});
```
apalaniuk, line 180: We probably want to have this event handler remove itself on being invoked, to ensure it doesn't outlive the test/invoke the `done` for this test in another context unintentionally, etc.

It could also be added to the rendered fixture, so it's cleaned up once the fixture is torn down. You'd want to assert that the event has `bubbles: true` and `composed: true` in that case (ie, consumers can handle it).

Another option could be to stub + spy `document.addEventListener`, and check that it was called for that event synchronously, then clean up the spy after the test completes.

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -75,7 +89,9 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 			'en': {
```
EvanLemieux, line 89: Maybe not for this PR but should these translations get pulled into an external file? I can imagine this file getting quite long once they are all filled in.

adamsk34: I agree that it should get pulled into an external file. `d2l-filter-dropdown-category.js` is another example of a file that does this and it's localization code should also probably move to an external file.

There are some places in this repo where it has been extracted to another file, example: `d2l-filter-dropdown-localize-behavior.js`. I feel like the best time to move it to an external file in when we start to fill it up with more translations.

apalaniuk: Definitely not for this PR, since this PR doesn't touch langterms. We can file a separate issue/story/whatever for that. (And more importantly IMO, getting this set up for automated translations)

apalaniuk: Definitely not for this PR, since that's increasing the scope. But more important than the file size to me is that this repo isn't part of our automated translations process, so separately, we should either do that (which almost certainly requires splitting out these files regardless of size), or requesting translations directly

apalaniuk (approved):

components/d2l-applied-filters/d2l-applied-filters.js
dbatiste: Which element is this selector being applied to?  Is it the subtle button?  The subtle button already contains host styles that honor the `hidden` attribute that is being applied.  In this case, the component will be scoped by shadowDOM so there isn't really a major concern of random things on the page overriding in this scope.

It is true that we try to avoid using `!important` for the specificity reason.  It basically trumps everything else, even inline styles.  It's basically the nuclear option.  Using it really tends to put us in a tight spot if we need to override the property.  For that reason, we only use it when it is absolutely necessary.

adamsk34: Yes this is for the subtle button. This PR is already merged but if you would like I can create a new PR to remove `!important`

dbatiste: I'd probably clean it up, but it's not a big deal.  Mostly worried about it being repeated.  Our goal should be to never use it, but I know we need to use it from time to time.

adamsk34: Ok, it wouldn't be much work for me to create a new PR. I'll do that now. Thanks for bringing it up.

PR #63, opened 2020-09-01
Use Lit linting config

svanherk:
More linting, updated to use the Lit linting config since it has a lot more rules and about half the repo is in Lit.

PR #64, opened 2020-09-01
Make the applied filter find a shadow-dom-nested filter-dropdown better

jstefaniuk-d2l:
The shadow-dom boundary was causing some issues for finding the dropdown filter wrapped in another component, this fixes that

Ideally this goes in after https://github.com/BrightspaceUILabs/facet-filter-sort/pull/63 so that it can have proper linting

### DIFF
--- a/components/d2l-applied-filters/d2l-applied-filters.js
+++ b/components/d2l-applied-filters/d2l-applied-filters.js
@@ -145,7 +145,8 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 				if (child.nodeName === DROPDOWN_NAME) {
 					return child;
 				}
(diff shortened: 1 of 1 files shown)


jstefaniuk-d2l:
Ideally this would be released ASAP I think, since I have another PR which will be waiting on it

svanherk:
> Ideally this would be released ASAP I think, since I have another PR which will be waiting on it, however let me know if there are any things to be done before a release

Ok, then I think the faster way would be to make a minor release now, rather than trying to rush in the rename and major bump.  There's that small rewrite due to the linting errors, but you could regression test that as part of this

jstefaniuk-d2l:
Tried it out with the changes (with master pulled in) locally and everything seemed to work like before 😄 

I'll merge this in and do a minor bump

PR #65, opened 2020-09-02
Move demo components to lit

svanherk:
Quickly moved the demo components over to Lit since they're fairly simple and low risk


demo/d2l-filter-dropdown/templated-filter-dropdown-demo.js
```diff
@@ -0,0 +1,193 @@
...
+import { html, LitElement } from 'lit-element';
+
+/**
+ * An example of a templated filter-dropdown.
+**/
+class TemplatedFilterDropdownFilter extends LitElement {
+	static get properties() {
+		return {
+			_filters: { type: Array },
+			_totalSelected: { type: Number }
```
dlockhart, line 13: If these aren't intended to be set by consumers (i.e. private), then you can add `attribute: false` to remove all the attribute binding stuff.

PR #69, opened 2020-09-29
Intrepid US120036 - Updates to work with new multi-select

marcsoiferman:
Still TODO:
- [x] Bump multi-select package
- [x] Fix tests after bump
- [x] Bump this version

Updates to allow this to work with the new multi-select component.  Moved the "Clear Filters" button into the mult-select component.

See https://github.com/BrightspaceUILabs/multi-select/pull/69

### DIFF
--- a/components/d2l-applied-filters/d2l-applied-filters.js
+++ b/components/d2l-applied-filters/d2l-applied-filters.js
@@ -134,7 +134,7 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 		const filters = this._selectedEntries && this._selectedEntries.length > 0 ?
 			html`<d2l-labs-multi-select-list
 				collapsable
-				shrinkwrap="true"
+				?shrinkwrap="${this.shrinkwrap}"
 				?show-clear-list="${this._selectedEntries.length >= CLEAR_FILTERS_THRESHOLD}"
 				clear-list-button-text="${this.localize('clearFilters')}"
 				@d2l-multi-select-list-clear-list-clicked="${this._clearFiltersClicked}"
--- a/components/d2l-applied-filters/d2l-applied-filters.js
+++ b/components/d2l-applied-filters/d2l-applied-filters.js
@@ -134,7 +134,7 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 		const filters = this._selectedEntries && this._selectedEntries.length > 0 ?
 			html`<d2l-labs-multi-select-list
 				collapsable
-				shrinkwrap?="true"
+				shrinkwrap="true"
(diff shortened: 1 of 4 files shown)


test/d2l-applied-filters/d2l-applied-filters.html
```diff
@@ -142,65 +142,79 @@
...
+						const buttons = filterList.shadowRoot.querySelectorAll('d2l-button-subtle');
+						buttons.forEach(a => {
+							if (a.getAttribute('text') === 'Clear filters' && a.getAttribute('class') !== 'hide') {
+								clearFilters = a;
+							}
+						});
 						clearFilters.click();
 					});
 				});
-				test('selecting a 4th filter & clicking clear filters removes filters & hides clear filters button', function(done) {
```
marcsoiferman, line 191: Due to a bug in the demo page, the items are still selected even though they've been cleared and the button sticks around invisible, so we don't have any easy way to detect if you can see it or not.

test/d2l-applied-filters/d2l-applied-filters.html
```diff
@@ -142,65 +142,79 @@
...
 					document.addEventListener('d2l-filter-dropdown-cleared', handleClearEvent);
 
 					options[1].click();
 
-					afterNextRender(clearFilters, () => {
+					afterNextRender(filterList, () => {
+						let clearFilters = undefined;
+						const buttons = filterList.shadowRoot.querySelectorAll('d2l-button-subtle');
+						buttons.forEach(a => {
+							if (a.getAttribute('text') === 'Clear filters' && a.getAttribute('class') !== 'hide') {
```
apalaniuk, line 192: I might prefer an attribute to a class for indicating hidden state (using the `hidden` attribute, maybe) as that feels more resilient to future changes/a public contract, but I do see that's what's being set today. For the class check, I think the best way to do that is `a.classList.contains('hide')`, since more classes may be added in the future

components/d2l-applied-filters/d2l-applied-filters.js
```diff
@@ -157,7 +161,6 @@ class D2lAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 				<span id="d2l-applied-filters-label" class="d2l-applied-filters-applied-filters-label d2l-body-compact">${this.localize('appliedFilters')}</span>
 				<div id="d2l-list-holder">
 					${filters}
-					<d2l-button-subtle id="d2l-clear-filters-button" text="${this.localize('clearFilters')}" ?hidden="${this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD}" @click="${this._clearFiltersClicked}"></d2l-button-subtle>
```
jstefaniuk-d2l, line 160: https://github.com/BrightspaceUILabs/facet-filter-sort/pull/69/files#diff-53571d5ed7be88c8a70d36ce8872776dR64 will also need removing

jstefaniuk-d2l (commented):
code LGTM

PR #72, opened 2020-10-09
Revert Shrinkwrap Support

marcsoiferman:
Reverting changes to pull shrinkwrap support out, since we had an unresolvable issue in Firefox with the multi select when used in combination with the filters.

Hoping we don't need to revert multi-select, since everything in there should be transparent to this component if we disable the support.


jstefaniuk-d2l (approved):

adamsk34:
## Testing

✅ passed

Tested using the demo.

### Browsers
- Chrome (v85)
- Firefox (v81)

### What I looked out for
- RTL and LTR
- List with no filters
- List with all filters applied
- Expanding and collapsing the list

PR #78, opened 2021-01-05
BREAKING CHANGE: US120575 - Rename Components to Labs

svanherk:
It's finally here!  Or at least, step 1 of 20...
This PR renames all the components, their files, their events, etc to include `labs`.  This will be a major version bump, and then I'll open a PR registering this repo with npm.  Then I'll open PRs for all the places using this in BSI, which will all need to merge at the same time.


dlockhart (approved):
LGTM

PR #80, opened 2021-01-06
fix: install web-component-analyzer in release only

svanherk:
Forgot to `npm i`, but to avoid that in release I'll just install the analyzer there.


dlockhart (approved):
Nice

PR #84, opened 2021-08-05
fix: awaiting updateComplete instead of requestUpdate()

dlockhart:
`requestUpdate` [no longer returns a Promise](https://lit.dev/docs/releases/upgrade/#litelement) in Lit 2.0, so instead you're supposed to `await elem.updateComplete`.

### DIFF
--- a/components/filter-dropdown/filter-dropdown-category.js
+++ b/components/filter-dropdown/filter-dropdown-category.js
@@ -155,7 +155,8 @@ class LabsFilterDropdownCategory extends LocalizeStaticMixin(TabPanelMixin(LitEl
 		const oldVal = this._categoryText;
 		if (oldVal !== val) {
 			this._categoryText = val;
-			this.requestUpdate().then(() => {
+			this.requestUpdate();
+			this.updateComplete.then(() => {
 				this._updateTabText(this._categoryText, this.selectedOptionCount);
 			});
 		}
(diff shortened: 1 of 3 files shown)


dlockhart:
The failure in Safari is interesting. It's failing because this exception is thrown:
> ResizeObserver loop completed with undelivered notifications

Safari only added support for ResizeObserver [in January 2020](https://webkit.org/blog/9997/resizeobserver-in-webkit/), which could explain why we weren't seeing this before... although there was a successful run two months ago.

The explanation for that error is this:
> If any changes are incurred during the callback, then layout happens again, but here, the system finds the shallowest at which depth a change occurred (measured in simple node depth from the root). Any changes that are related to something deeper down in the tree are delivered at once, while any that are not are queued up and delivered in the next frame, and an error message will be sent to the Web Inspector console: (ResizeObserver loop completed with undelivered notifications).

Digging deeper:

- This originates from [hierarchical view's ResizeObserver](https://github.com/BrightspaceUI/core/blob/d62d4b0c391d5793c4db113d491782f07c2e4bf5/components/hierarchical-view/hierarchical-view-mixin.js#L519) which when triggered just [fires a d2l-hierarchical-view-resize event](https://github.com/BrightspaceUI/core/blob/d62d4b0c391d5793c4db113d491782f07c2e4bf5/components/hierarchical-view/hierarchical-view-mixin.js#L311)
- That in turn is listened to by menu, which [fires its own d2l-menu-resize event](https://github.com/BrightspaceUI/core/blob/d62d4b0c391d5793c4db113d491782f07c2e4bf5/components/menu/menu.js#L350)
- Lastly, dropdown-tabs is listening for that which [re-positions itself](https://github.com/BrightspaceUI/core/blob/d62d4b0c391d5793c4db113d491782f07c2e4bf5/components/dropdown/dropdown-tabs.js#L55) and completes the cycle and triggers a new resize observation on the original hierarchical-view.

I can't figure out why Safari is complaining and the other browsers aren't... everything seems to be happening in the same sequence.

dlockhart:
Raised this [as a defect on core](https://github.com/BrightspaceUI/core/issues/1575) to tackle later, skipping the test for now.

svanherk (approved):

PR #87, opened 2021-11-10
Pull multi-select from npm

BhamelinD2L:
Switches the multi-select reference from github to instead use the npm registry through https://www.npmjs.com/package/@brightspace-ui-labs/multi-select

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "dependencies": {
     "@brightspace-ui/core": "^1",
     "d2l-localize-behavior": "BrightspaceUI/localize-behavior#semver:^2",
-    "@brightspace-ui-labs/multi-select": "BrightspaceUILabs/multi-select#semver:^4",
+    "@brightspace-ui-labs/multi-select": "^4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
   "dependencies": {
     "@brightspace-ui/core": "^1",
     "d2l-localize-behavior": "BrightspaceUI/localize-behavior#semver:^2",
-    "@brightspace-ui-labs/multi-select": "BrightspaceUILabs/multi-select#semver:^4",
+    "@brightspace-ui-labs/multi-select": "^4.2.4",
```
dlockhart, line 41: Could you just make this `^4`? Typically we try to avoid including minor versions unless it specifically needs a feature in them. BSI will always take the latest version due to the `^` anyway.
```suggestion
    "@brightspace-ui-labs/multi-select": "^4",
```

BhamelinD2L:
I originally made the PR at ^4, but the optional chaining actually was an issue in multi-select at an earlier version. I assumed ^4 would still get the latest but wanted to double check, thanks. I've switched all PRs to ^4.

dlockhart (approved):

PR #100, opened 2022-02-04
Lit 2 Upgrade: backwards-compatible changes

svanherk:
This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
  - Only usage is in a Polymer component
- [x] audit usages of `unsafeHTML` to ensure `null`, `undefined` inputs are handled as expected
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
- [x] audit `ifDefined` usages for null parameters
  - Usage audited, new behavior makes things better!

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.

### DIFF
--- a/components/applied-filters/applied-core-filters.js
+++ b/components/applied-filters/applied-core-filters.js
@@ -1,7 +1,7 @@
 import '@brightspace-ui/core/components/button/button-subtle.js';
 import '@brightspace-ui-labs/multi-select/multi-select-list.js';
 import '@brightspace-ui-labs/multi-select/multi-select-list-item.js';
-import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { css, html, LitElement } from 'lit-element';
 import { announce } from '@brightspace-ui/core/helpers/announce.js';
 import { bodyCompactStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { IdSubscriberController } from '@brightspace-ui/core/controllers/subscriber/subscriberControllers.js';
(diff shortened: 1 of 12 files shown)


demo/filter-dropdown/filter-dropdown.html
```diff
@@ -22,77 +22,73 @@
 			<h2>d2l-labs-filter-dropdown with dom-repeat data and event hookups</h2>
 
 			<d2l-demo-snippet>
-				<template>
```
svanherk, line 25: I don't think the `template`s here needed to be removed because these are demo files for Polymer components lol.  But I don't think they're doing anything useful.

dlockhart (approved):
LGTM

components/applied-filters/applied-core-filters.js
```diff
@@ -1,7 +1,7 @@
 import '@brightspace-ui/core/components/button/button-subtle.js';
 import '@brightspace-ui-labs/multi-select/multi-select-list.js';
 import '@brightspace-ui-labs/multi-select/multi-select-list-item.js';
-import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { css, html, LitElement } from 'lit-element';
```
dlockhart, line 4: I made this change in a couple of my PRs and then realized after it's not really necessary. These ultimately will end up getting switched over to `import { css, html LitElement } from 'lit'` anyway.

annabelsegalld2l (approved):

PR #101, opened 2022-02-07
US134327 - Remove text for portfolio header changes

EvanLemieux:
## Description
Changes proposed are to satisfy a design change in folio. 

image

image

### DIFF
--- a/components/applied-filters/applied-filters.js
+++ b/components/applied-filters/applied-filters.js
@@ -158,11 +158,10 @@ class D2lLabsAppliedFilters extends RtlMixin(LocalizeStaticMixin(LitElement)) {
 					</d2l-labs-multi-select-list-item>
 				`)}
 			</d2l-labs-multi-select-list>`
-			: html`<span class="d2l-labs-applied-filters-none-label d2l-body-compact">${this.localize('noActiveFilters')}</span>`;
+			: null;
 
 		return html`
 			<div class="d2l-labs-applied-filters-wrapper">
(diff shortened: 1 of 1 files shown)


jameswklassen:
Are the langterms still used for accessibility purposes? If not we can probably remove them? 🤷

EvanLemieux:
> Are the langterms still used for accessibility purposes? If not we can probably remove them? 🤷


> Are the langterms still used for accessibility purposes? If not we can probably remove them? 🤷

They're used in the other component still, so I didn't remove them

svanherk:
Just checking - this is something we want changed everywhere these are used?  Insights dashboards also use these tags.

EvanLemieux:
> Just checking - this is something we want changed everywhere these are used? Insights dashboards also use these tags.

Code search showed only a test using this component, but if they do use it then no. This was a change intended for only folio.

svanherk:
> Code search showed only a test using this component, but if they do use it then no. This was a change intended for only folio.
Edit: Further investigation leads me to believe they have created their own implementation? https://github.com/Brightspace/insights-assessment-quality-dashboard/blob/master/components/applied-filters.js

Oh wow, ok that's good to know... I wasn't aware of that.  Official tag and tag group components (and then filter tags) is coming soon so we'll need to learn what functionality they needed that wasn't available to see if we can incorporate it.

But anyways, looks like you're right and that doesn't affect what you're doing here.  Ignore me!

jameswklassen (approved):

PR #105, opened 2022-03-25
Add Serge

dlockhart:
This merges all the various ways localization was set up in this repo into a single localization file with a Lit mixin and a Polymer behaviour.

That allows for Serge to be set up, which I'll connect once this merges.


package.json
```diff
@@ -10,14 +10,15 @@
...
-    "/components"
+    "/components",
+    "/lang"
   ],
   "scripts": {
-    "lint": "npm run lint:eslint && npm run lint:lit && npm run lint:polymer && npm run lint:style",
+    "lint": "npm run lint:eslint && npm run lint:lit && npm run lint:style",
     "lint:eslint": "eslint . --ext .js,.html",
     "lint:lit": "lit-analyzer \"{components, test, demo}/**/*.js\" --strict --rules.no-unknown-tag-name off",
-    "lint:polymer": "polymer lint -i \"test/**/*.js\" \"demo/**/*.js\" \"components/**/*.js\"",
```
dlockhart, line 20: Polymer linting hated our Lit dynamic import, so it was easier to just disconnect it at this point.

test/applied-filters/applied-core-filters.test.js
```diff
@@ -57,6 +57,8 @@ describe('d2l-labs-applied-core-filters', () => {
 			const elem = await fixture(basic);
 			filter = elem.querySelector('d2l-filter');
 			appliedCoreFilters = elem.querySelector('d2l-labs-applied-core-filters');
+			await filter.updateComplete;
+			await appliedCoreFilters.updateComplete;
```
dlockhart, line 61: These are now deferred a bit while the localization loads -- if they were the root of the fixture it would automatically await them. I may need to add more of these if tests fail in Sauce.

dlockhart:
I'm going to have to circle back to these tests on Monday. I thought converting to Playwright would help with the Safari problems, but it seems worse -- can't repro any of them locally.

test/filter-dropdown/filter-dropdown-category.test.js
```diff
@@ -28,7 +28,6 @@ describe('d2l-labs-filter-dropdown-category', () => {
 
 	beforeEach(async() => {
 		container = await fixture(basic);
-		await container.updateComplete;
```
dlockhart, line 31: Not necessary since `fixture` already awaits it.

test/filter-dropdown/filter-dropdown-category.test.js
```diff
@@ -38,14 +37,18 @@ describe('d2l-labs-filter-dropdown-category', () => {
 	it('instantiating the element works', () => {
 		expect('d2l-labs-filter-dropdown-category').to.equal(categories[0].tagName.toLowerCase());
 	});
-	it('attributes are set correctly (including d2l-tab-panel attributes)', () => {
+	it('attributes are set correctly (including d2l-tab-panel attributes)', async() => {
+		await categories[0].updateComplete;
+		await categories[1].updateComplete;
```
dlockhart, line 42: A nicer solution would have been to move this up into `beforeEach`, but at least one of the tests ('changing the category tab triggers the d2l-labs-filter-dropdown-category-selected event') immediately `await`s an event that's already fired if we await `updateComplete`. These probably need a rethink at some point.

test/filter-dropdown/filter-dropdown-category.test.js
```diff
@@ -85,7 +90,7 @@ describe('d2l-labs-filter-dropdown-category', () => {
 		let e = await oneEvent(container, 'd2l-labs-filter-dropdown-category-selected');
 		expect(e.detail.categoryKey).to.equal('1');
 
-		categories[1].selected = true;
+		setTimeout(() => categories[1].selected = true);
```
dlockhart, line 93: Just for safety in case the event is fired synchronously from the setter.

dlockhart (commented):
Ok I think this is finally ready for review!

components/localize-mixin.js
```diff
@@ -0,0 +1,11 @@
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
+
+export const FilterLocalizeMixin = superclass => class extends LocalizeDynamicMixin(superclass) {
```
svanherk, line 3: Minor: Seems like this is all langterms (not just filter langterms) so something like just `LocalizeMixin` would probably be better but the name doesn't _really_ matter....

dlockhart: Agreed yeah... I didn't want to call it `LocalizeMixin` because that's what the "core" one is called... I guess `FacetFilterSortLocalizeMixin` is the closest but that's a mouthful.

lang/cy.js
```diff
@@ -0,0 +1,23 @@
...
+	"clearFilters": "Clirio hidlyddion",
+	"allFiltersRemoved": "Tynnwyd pob hidlydd",
+	"categoryTitleMultiple": "{title} ({numSelected})",
+	"searchBy": "Chwilio yn ôl {category}",
+	"clear": "Clirio",
+	"filter": "Hidlo",
+	"filterBy": "Hidlo yn ôl",
+	"filterMultiple": "Hidlo: {numOptions} Hidlyddion",
+	"filterSingle": "Hidlo: 1 Hidlydd",
+	"more": "mwy",
```
svanherk, line 16: I wonder if the More langterms (which I'm guessing you got from `core`) should be capitalized?

lang/da.js
```diff
@@ -0,0 +1,21 @@
...
+	"clear": "Ryd",
+	"filter": "Filtrér",
+	"filterBy": "Filtrér efter",
+	"filterMultiple": "Filtrér: {numOptions} filtre",
+	"filterSingle": "Filtrér: 1 filter",
+	"more": "flere",
+	"searchPagedResults": "{rangeStart}-{rangeEnd} af {totalCount} resultater",
+	"searchPagedResultsForQuery": "{rangeStart}-{rangeEnd} af {totalCount} resultater for \"{query}\"",
+	"searchResults": "{totalCount} resultater",
+	"searchResultsForQuery": "{totalCount} resultater for \"{query}\""
```
svanherk, line 20: Should this and `de.js` have `sort` and `sortWithOption` with English terms?

dlockhart: Good catch -- added!

lang/es-es.js
```diff
@@ -0,0 +1,23 @@
+/* eslint quotes: 0 */
+
+export default {
+	"appliedFilters": "Filtros aplicados:",
+	"noActiveFilters": "No hay filtros activos",
```
svanherk, line 5: Wondering if this specific term is backwards between `es` and `es-es`... this was the langterm from `es-mx`, and we rewrite from `es-mx` to `es`.  BUT we don't want to put the `es-es` term for `es`... so maybe both should be `Sin filtros activos`?

lang/cy.js
svanherk: Also noticed this in `da.js` and `de.js`

lang/es-es.js
dlockhart: Yeah I agree -- it's hard to reconcile but before `es-es` would have resolved to `es` so it makes sense I think for them to both be the same. Updated.

lang/cy.js
dlockhart: Right -- updated.

svanherk (approved):
Thanks!  Looks like this was just as painful as I guessed it would be 🙈

PR #114, opened 2022-07-13
add filter option for sort-by-dropdown

JamieTran:
https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F636678076593

Will be used in https://github.com/Brightspace/consistent-evaluation/pull/1114

- Added an attribute (`?is-filter`) to determine whether to display `Filter:` or `Sort:` when using the `d2l-labs-sort-by-dropdown`.

Consistent-evaluation (quizzing) needs a filter-dropdown component where only one selection can be made. The exact functionality of `d2l-labs-sort-by-dropdown`.  I considered making the text editable, but being it is in `facet-filter-sort`, I just left it with those two options.

question-filtering

### DIFF
--- a/components/sort-by-dropdown/sort-by-dropdown.js
+++ b/components/sort-by-dropdown/sort-by-dropdown.js
@@ -45,8 +45,8 @@ class LabsSortByDropdown extends mixinBehaviors(
 				value: '',
 				observer: function(selection) {
 					this._selectedOptionText = this.compact
-						? this.localize('sort')
-						: this.localize('sortWithOption', 'option', selection);
+						? this.isFilter ? this.localize('filter') : this.localize('sort')
+						: this.isFilter ? this.localize('filterWithOption', 'option', selection) : this.localize('sortWithOption', 'option', selection);
 				}
 			},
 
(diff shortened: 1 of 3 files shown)


maddiegabriel (approved):
lgtm!

svanherk:
Yeah, I was talking to @geurts about this.  The feedback we're getting is that the `single-selection` version of the `filter` feels too heavy for simple situations like the one above, and we're not sure what to do about it yet.  I also don't love repurposing this sort component, but I'm not sure if its better to do that, do a one-off in the project for now, or use the `single-selection` filter and we'll just try to adjust/improve it

JamieTran:
> Yeah, I was talking to @geurts about this. The feedback we're getting is that the `single-selection` version of the `filter` feels too heavy for simple situations like the one above, and we're not sure what to do about it yet. I also don't love repurposing this sort component, but I'm not sure if its better to do that, do a one-off in the project for now, or use the `single-selection` filter and we'll just try to adjust/improve it

Oh this is perfect, initially looking at the `core filter` component I didn't see the attribute to allow `single-selection` and thought it was supposed to be used for applying multiple filters, this is much better than the change outlined in this PR 😃

PR #116, opened 2023-08-21
US156707 - Cleanup unused components

svanherk:
`d2l-labs-search-facets` and `d2l-labs-search-results-count` are not used (and have not been for a long time).  Confirmed with Jeff that we should just remove them and not migrate them.

I'll put up a PR in `documentation` to clean up the page referencing them.


dlockhart (approved):
Excellent!

PR #117, opened 2023-08-21
Some more cleanup while here

svanherk:
Only did the easy stuff.


dlockhart (approved):
One less place `console.warn`-ing about the Lit stuff!

PR #118, opened 2023-09-20
Use dynamic lang import

bearfriend:
Polymer components can now use dynamic imports for language files. I also switched to the new @brightspace-ui/testing library to help test the changes more easily.


dlockhart (commented):
CI workflow appears to still be referencing the config file and the install step can be tweaked.

dlockhart (approved):

PR #121, opened 2024-01-23
Update `_text` when selected option text changes

bearfriend:
Because `localize` is asynchronous, option `text` may render after `_selectedOptionText` is computed (from `_text`), so notify when it changes.

### DIFF
--- a/components/sort-by-dropdown/sort-by-dropdown-option.js
+++ b/components/sort-by-dropdown/sort-by-dropdown-option.js
@@ -20,6 +20,16 @@ class LabsSortByDropdownOption extends RtlMixin(MenuItemRadioMixin(LitElement)) 
 		`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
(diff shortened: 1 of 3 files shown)


components/sort-by-dropdown/sort-by-dropdown-option.js
```diff
@@ -20,6 +20,16 @@ class LabsSortByDropdownOption extends RtlMixin(MenuItemRadioMixin(LitElement))
 		`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('text') && this.selected) {
+			this.dispatchEvent(new CustomEvent('d2l-labs-sort-by-dropdown-option-selected-text-change', {
+				detail: { text: this.text },
+				bubbles: true }
+			));
```
svanherk, line 29:
```suggestion
				bubbles: true
			}));
```

svanherk (approved):

PR #124, opened 2024-11-14
fix: Replace multi-select-list with tag-list

GZolla:
Replace deprecated component `multi-select-list` in favor of the core component `d2l-tag-list`

### DIFF
--- a/components/applied-filters/applied-filters.js
+++ b/components/applied-filters/applied-filters.js
@@ -1,11 +1,8 @@
-import '@brightspace-ui/core/components/button/button-subtle.js';
 import '@brightspace-ui/core/components/tag-list/tag-list.js';
 import '@brightspace-ui/core/components/tag-list/tag-list-item.js';
-import { css, html, LitElement } from 'lit';
+import { css, html, LitElement, nothing } from 'lit';
 import { bodyCompactStyles } from '@brightspace-ui/core/components/typography/styles.js';
-import { FilterLocalizeMixin } from '../localize-mixin.js';
 import { getComposedChildren } from '@brightspace-ui/core/helpers/dom.js';
-import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 
 const DROPDOWN_NAME = 'D2L-LABS-FILTER-DROPDOWN';
 const DROPDOWN_CATEGORY_NAME = 'D2L-LABS-FILTER-DROPDOWN-CATEGORY';
@@ -14,7 +11,7 @@ const DROPDOWN_OPTION_NAME = 'D2L-LABS-FILTER-DROPDOWN-OPTION';
 /**
  * A tag-list allowing the user to see (and remove) the currently applied filters.
  */
-class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
+class D2lLabsAppliedFilters extends LitElement {
 	static get properties() {
 		return {
 			/**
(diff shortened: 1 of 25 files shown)


components/applied-filters/applied-filters.js
```diff
@@ -103,27 +96,27 @@ class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
 
 	render() {
 		const filters = this._selectedEntries && this._selectedEntries.length > 0 ?
-			html`<d2l-labs-multi-select-list
-				collapsable
+			html`<d2l-tag-list
+				clearable
 				aria-labelledby="d2l-labs-applied-filters-label"
```
dlockhart, line 101: This `aria-labelledby` is pointing at an element that doesn't appear to exist? Also `d2l-tag-list` doesn't support such an attribute, and requires a `description` instead.

I guess `d2l-labs-applied-filters`'s `labelText` property should go there? It doesn't appear to be used currently.

components/applied-filters/applied-filters.js
```diff
@@ -103,27 +96,27 @@ class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
...
+				@d2l-tag-list-clear="${this._clearFiltersClicked}"
 			>
 				${(this._selectedEntries || []).map((x, index) => html`
-					<d2l-labs-multi-select-list-item
+					<d2l-tag-list-item
 						text="${x.text}"
 						deletable
```
dlockhart, line 107: `deletable` isn't a thing on `d2l-tag-list-item`, should this be `clearable`?

components/applied-filters/applied-filters.js
```diff
@@ -103,27 +96,27 @@ class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
...
-					</d2l-labs-multi-select-list-item>
+					</d2l-tag-list-item>
 				`)}
-			</d2l-labs-multi-select-list>`
+			</d2l-tag-list>`
 			: null;
 
 		return html`
 			<div class="d2l-labs-applied-filters-wrapper">
 				<div id="d2l-list-holder">
```
dlockhart, line 118: I feel like we should be able to get rid of these two `div`s and their CSS.

components/applied-filters/applied-filters.js
```diff
@@ -103,27 +96,27 @@ class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
...
 					${filters}
-					<d2l-button-subtle id="d2l-clear-filters-button" text="${this.localize('clearFilters')}" ?hidden="${this._selectedEntries.length < CLEAR_FILTERS_THRESHOLD}" @click="${this._clearFiltersClicked}"></d2l-button-subtle>
```
dlockhart, line 126: The `button-subtle.js` import can go away now.

components/applied-filters/applied-filters.js
```diff
@@ -66,11 +64,6 @@ class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
 				padding-top: 0.3rem;
 			}
```
dlockhart, line 66: Not part of your change, but it looks like `.d2l-labs-applied-filters-label` can go away too, and it was the only thing requiring the `RtlMixin`.

components/applied-filters/applied-filters.js
```diff
@@ -39,7 +37,7 @@ class D2lLabsAppliedFilters extends RtlMixin(FilterLocalizeMixin(LitElement)) {
 				display: block;
 			}
 
-			d2l-labs-multi-select-list {
+			d2l-tag-list {
```
dlockhart, line 40: This CSS probably isn't needed anymore.

test/applied-filters/applied-filters.test.js
```diff
@@ -2,7 +2,7 @@ import '../../components/filter-dropdown/filter-dropdown.js';
 import '../../components/filter-dropdown/filter-dropdown-category.js';
 import '../../components/filter-dropdown/filter-dropdown-option.js';
 import '../../components/applied-filters/applied-filters.js';
-import { expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { clickElem, expect, fixture, html, oneEvent, sendKeysElem, waitUntil } from '@brightspace-ui/testing';
 import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
```
dlockhart, line 6: Would you be able to convert the other 2 files using `@open-wc/testing` over as well? Looks like they were missed when this repo was migrated to `@brightspace-ui/testing`.

```suggestion
import { clickElem, expect, fixture, html, oneEvent, runConstructor, sendKeysElem, waitUntil } from '@brightspace-ui/testing';
```

dlockhart (commented):
Nice job! I didn't think this would be this easy. Just a few more things to clean up I think.

dlockhart (approved):
LGTM! 🎉

PR #137, opened 2026-05-06
feat: filter-dropdown-category: no longer listen for d2l-tab-panel-selected event

margaree:
[GAUD-9869](https://desire2learn.atlassian.net/browse/GAUD-9869)

This switch is a bit more complex since it's using tabPanelMixin. I'll leave comments.

We are deprecating `d2l-tab-panel-selected` event so this switches to listening for `d2l-tab-selected` instead.

[GAUD-9869]: https://desire2learn.atlassian.net/browse/GAUD-9869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,10 +126,10 @@ Note that for the header and opener text overrides, if the terms are to reflect 
 
 - `d2l-labs-filter-dropdown-close`: Fired when the filter dropdown is closed.
 - `d2l-labs-filter-dropdown-cleared`: Fired when the clear button or the clear filters button is pressed to clear all filters.
+- `d2l-labs-filter-dropdown-category-selected`: Fired when a filter tab is selected.
 
 `d2l-labs-filter-dropdown-category`:
 
-- `d2l-labs-filter-dropdown-category-selected`: Fired when a filter tab is selected.
 - `d2l-labs-filter-dropdown-category-searched`: Fired when a filter category is searched.
 - `d2l-labs-filter-dropdown-option-change`: Fired when a filter option is selected.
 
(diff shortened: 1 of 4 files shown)


margaree:
Since I can't add inline comments: 

- https://github.com/BrightspaceUILabs/facet-filter-sort/pull/137/changes#diff-b224a9aa05739e381ea76b1ef1f1af21151e07d9518f281d54f8e33963ed1d41L148 - no longer needed
- https://github.com/BrightspaceUILabs/facet-filter-sort/pull/137/changes#diff-b224a9aa05739e381ea76b1ef1f1af21151e07d9518f281d54f8e33963ed1d41R182 - This does change where the event is dispatched from. It's only listened for in one place which wont be impacted by it moving and just looks for `categoryKey` ([here](https://search.d2l.dev/xref/Brightspace/folio-ui/src/web-components/categories-filter-menu/index.js?r=2e843030a93e5823ebd51bd67b8ee5034c9dc899#77); it fires from within the custom component).
But yeah it isn't necessarily the pattern we generally fire so let me know if concerns and I can see try to move it into -category, I think that's just a more complex change.
- https://github.com/BrightspaceUILabs/facet-filter-sort/pull/137/changes#diff-82673f470730b1e19c605b9eb0590d0b41e11d1f90b6a07e2da8ecc8ad60792cR9 - I checked to make sure other places weren't wrapping this component in `d2l-tabs` since that would no longer work and found none

dlockhart (approved):

